PR #3, opened 2017-06-22
Support for finding Xamarin.Android location with only VS 2017 installed

jonathanpeppers:
- Included a unit test project with smoke tests (I just used 2.x NUnit). These are passing on my Windows machine with VS 2017.
~~- Uses [vswhere](https://github.com/Microsoft/vswhere) to locate a VS 2017 installation that includes Xamarin. I think this is the preferred method for locating VS 2017 now.~~
- Looks for VS 2017 in known locations

### DIFF
--- a/src/Xamarin.Android.Tools/Sdks/MonoDroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools/Sdks/MonoDroidSdkBase.cs
@@ -108,8 +108,7 @@ namespace Xamarin.Android.Tools
 			return !string.IsNullOrWhiteSpace (loc) &&
 				(File.Exists (Path.Combine (loc, DebugRuntime)) ||    // Normal/expected
 				 File.Exists (Path.Combine (loc, ClassParseExe)) ||    // Normal/expected
-					File.Exists (Path.Combine (loc, "Ionic.Zip.dll")) || // Wrench builds
-					File.Exists (Path.Combine (loc, "Xamarin.Android.Common.targets"))); //VS on Windows
+				 File.Exists (Path.Combine (loc, "Xamarin.Android.Common.targets"))); //VS on Windows
 		}
 
 		protected static bool ValidateFramework (string loc)
--- a/src/Xamarin.Android.Tools/Sdks/MonoDroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools/Sdks/MonoDroidSdkBase.cs
@@ -108,7 +108,8 @@ namespace Xamarin.Android.Tools
 			return !string.IsNullOrWhiteSpace (loc) &&
 				(File.Exists (Path.Combine (loc, DebugRuntime)) ||    // Normal/expected
 				 File.Exists (Path.Combine (loc, ClassParseExe)) ||    // Normal/expected
-					File.Exists (Path.Combine (loc, "Ionic.Zip.dll")));  // Wrench builds
+					File.Exists (Path.Combine (loc, "Ionic.Zip.dll")) || // Wrench builds
+					File.Exists (Path.Combine (loc, "Xamarin.Android.Common.targets"))); //VS on Windows
 		}
 
 		protected static bool ValidateFramework (string loc)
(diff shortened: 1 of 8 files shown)


jonpryor:
I have one conceptual problem with this: there can be *multiple* Visual Studio installations on a single machine.

For example, I have VS 2017 Enterprise and Community installed. When I run `vswhere`, I get:

```
> C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise>..\..\Installer\vswhere -latest -requires Component.Xamarin -property installationPath
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community
```

i.e. it returns the product I most recently installed, because of `-latest`. If I *remove* `-latest`, I get *all* installations:

```
> C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise>..\..\Installer\vswhere -requires Component.Xamarin -property installationPath
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise
```

The question is thus: how should `xamarin-android-tools` know which one is the appropriate and relevant version to use?

I am not sure that can be answered.

jonathanpeppers:
@dellis1972 @jonpryor 

I switched to probing directories. I tried @grendle's code from the installer, but it didn't include support for 2017. I also noticed they have removed the registry key in 2017 in favor of vswhere...

I think this solution is the most reliable for now.

src/Xamarin.Android.Tools/Sdks/MonoDroidSdkBase.cs
```diff
@@ -108,7 +108,8 @@ protected static bool ValidateRuntime (string loc)
 			return !string.IsNullOrWhiteSpace (loc) &&
 				(File.Exists (Path.Combine (loc, DebugRuntime)) ||    // Normal/expected
 				 File.Exists (Path.Combine (loc, ClassParseExe)) ||    // Normal/expected
-					File.Exists (Path.Combine (loc, "Ionic.Zip.dll")));  // Wrench builds
+					File.Exists (Path.Combine (loc, "Ionic.Zip.dll")) || // Wrench builds
```
jonpryor: We should remove this check, as we shouldn't even be using `Ionic.Zip.dll` anymore!

src/Xamarin.Android.Tools/Sdks/MonoDroidSdkWindows.cs
```diff
@@ -7,17 +7,33 @@ namespace Xamarin.Android.Tools
...
+
 		protected override string FindRuntime ()
 		{
 			string monoAndroidPath = Environment.GetEnvironmentVariable ("MONO_ANDROID_PATH");
 			if (!string.IsNullOrEmpty (monoAndroidPath)) {
 				if (Directory.Exists (monoAndroidPath) && ValidateRuntime (monoAndroidPath))
 					return monoAndroidPath;
 			}
+
 			string xamarinSdk = Path.Combine (OS.ProgramFilesX86, "MSBuild", "Xamarin", "Android");
```
jonpryor: There's another problem here: `xamarinSdk` is the VS 2015 install location. `VisualStudioPaths` contains VS 2017 locations.

It is possible to have *both* VS 2015 and VS 2017 installed.

Should we really prefer the VS 2015 install location over the VS 2017 install location, if both are installed? I think we should check for VS 2015 *after* VS 2017 locations.

Additionally, this could be cleaned up by "moving" `xamarinSdk` into `VisualStudioPaths`, so that they're treated uniformly.

src/Xamarin.Android.Tools/Sdks/MonoDroidSdkWindows.cs
```diff
@@ -7,17 +7,33 @@ namespace Xamarin.Android.Tools
...
-				? xamarinSdk
-					: OS.ProgramFilesX86 + @"\MSBuild\Novell";
+			if (Directory.Exists(xamarinSdk))
+				return xamarinSdk;
+
+			foreach (var vsPath in VisualStudioPaths) {
+				if (string.IsNullOrEmpty(vsPath))
+					continue;
+				xamarinSdk = Path.Combine(vsPath, "MSBuild", "Xamarin", "Android");
+				if (Directory.Exists(xamarinSdk) && ValidateRuntime(xamarinSdk))
```
jonpryor: Coding convention: space ` ` before `(`.

PR #5, opened 2017-09-07
Restructure the project for consistency with xamarin-android

jonpryor:
`Xamarin.Android.sln` is now a toplevel file, so it's easier to see
without navigating into the `src` directory.

Add a `Makefile` with three targets of consequence:

  * `make prepare`: Prepare the project for building, e.g.
    restore NuGet packages.
  * `make all`: Build the project.
  * `make run-all-tests`: Run the NUnit tests.

Move unit tests for `Xamarin.Android.Tools` into
`src/Xamarin.Android.Tools/Tests` (child, not a peer).

Rename the unit test assembly to `Xamarin.Android.Tools-Tests.dll`,
and build into `bin/Test$(Configuration)`.

Update to use NUnit 3.2.1 packages.

Update `README.md` with build instructions.

Add a `LICENSE` file.

Rename `Xamarin.Android.Tools.dll` to
`Xamarin.Android.Tools.AndroidSdk.dll`. Xamarin.Android already has
multiple `Xamarin.Android.Tools.*.dll` assemblies, e.g.
`Xamarin.Android.Tools.BootstrapTasks.dll`. Naming-wise,
`Xamarin.Android.Tools.dll` sounds like it should be *core* to *all*
related assemblies, but it isn't, and won't be. Instead, its core
responsibility is around the Android SDK, so name it accordingly.


Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+prepare:
+	nuget restore Xamarin.Android.Tools.sln
+
+run-all-tests: run-nunit-tests
+
+# $(call RUN_NUNIT_TEST,filename,log-lref?)
+define RUN_NUNIT_TEST
+	MONO_TRACE_LISTENER=Console.Out \
+	$(RUNTIME) --runtime=v4.0.0 \
```
dellis1972: should this not just be `v4.0` ? Not sure what windows will do with `v4.0.0`

jonpryor: It's different/worse from that: `--runtime` is a _mono_ option, and if mono is not used, there will thus be no command _at all_, which is bad.

On the ¯\_(ツ)_/¯ side, this is what xamarin-android does...

I'll remove the `--runtime` option. It might not be needed anymore now that I'm using NUnit 3.2.1.

dellis1972: I think the newer NUnit does a better job at detecting mono.

dellis1972: I tend to use the --framework option though..

PR #14, opened 2017-12-04
Add members for legacy compatibility

jonpryor:
The internal `androidtools` repo needs to be updated to use the
`xamarin-android-tools` versions of these types instead of its (older,
original) versions of these types, and in order to do so we need to
reintroduce some `AndroidSdkInfo` members which were removed.

Update `AndroidAppManifest` so that `androidtools` functionality can
be retained and reimplemented against the new `AndroidAppManifest`.


src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -116,5 +116,28 @@ public bool IsPlatformInstalled (int apiLevel)
 		public string AndroidNdkHostPlatform {
 			get { return sdk.NdkHostPlatform; }
 		}
+
+		public IEnumerable<string> AllAndroidSdkPaths {
```
dellis1972: should these be Obsoleted so that we are at least reminded we need to stop using them at some point?

PR #20, opened 2018-02-20
[windows] fallback path for Android SDK

jonathanpeppers:
In cases where `AndroidSdkInfo` is used outside of Visual Studio, such
as Embeddinator-4000, a new fallback path for the Android SDK is needed
on Windows.

At some point in the past, Visual Studio (or Xamarin) Installer
installed to:

    C:\Program Files (x86)\Android\android-sdk-windows

Now it is installed to (VS 2017):

    C:\Program Files (x86)\Android\android-sdk

Helps fix https://github.com/mono/Embeddinator-4000/issues/599 

Changes:
- Added another fallback path for the Android SDK, mirrored what I did
in the past for VS 2017 and the Android NDK
- Added .vs/ to the .gitignore for Windows developers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ bin
 **/obj
 packages/
 TestResult-*.xml
+.vs/
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -89,13 +89,17 @@ namespace Xamarin.Android.Tools
 			// Check some hardcoded paths for good measure
 			var xamarin_private = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.LocalApplicationData), "Xamarin", "MonoAndroid", "android-sdk-windows");
 			var android_default = Path.Combine (OS.ProgramFilesX86, "Android", "android-sdk-windows");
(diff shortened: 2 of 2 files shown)


tritao:
👍

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
```diff
@@ -15,5 +15,18 @@ public void Constructor_NullLogger ()
 			Action<TraceLevel, string> logger = null;
 			Assert.Throws<ArgumentNullException> (() => new AndroidSdkInfo (logger));
 		}
+
+		[Test]
+		public void AndroidSdkAndNdk_Exist ()
```
jonpryor: I'm not sure how I feel about this test. All it's really asserting is that *one* of the "well known" Android SDK directories exists on the machine running the unit tests.  This doesn't actually let us know that the `vs_2017_default` location added as part of this patch is one of them...nor could it. (How could it?)

Additionally, asserting that the NDK path exists is wrong, as the NDK is *optional*; see the `AndroidSdkInfo` constructor, in which we only assert that the Android SDK and Java JDK paths are valid.

jonathanpeppers:
@jonpryor that test was only helpful to diagnose on my Windows machine. I couldn't think of a good way to _really_ test this.

I just removed the test for now.

PR #24, opened 2018-04-06
Expose AllAndroidSdks as VS4Mac Installer depends on it

daserge:
It is needed to allow user to select between existing/new SDK locations


dellis1972:
Merging for now. If we need a better way of exposing this information then we should talk about that when @jonpryor gets back.

PR #25, opened 2018-04-16
[Xamarin.Android.Tools.AndroidSdk] Unstable framework versions are used

dellis1972:
We had a bug on our logic to calculate the `MaxStableVersion`.
The first time through the `LoadVersions` method `MaxStableVersion`
would ALWAYS be `null`. As a result regardless of whether the
first version was stable or not it would be assigned to the
property. This would then cause build failures later down the
build process because we were using an unstable api.

So the fix it to completely ignore unstable apis. This way the
first STABLE api we find will be the first value we assign.


jonpryor:
I'd also suggest a test which does:

```csharp
var versions = new AndroidVersions (
    new[]{new AndroidVersion (100, "100", "100-Test", "Z", false)});
Assert.IsNull (versions.MaxStableVersion);
Assert.IsNull (versions.MinStableVersion);
```

PR #27, opened 2018-05-22
Fix ~/.config/xbuild/* ownership on creation when under sudo

grendello:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/597752

Whenever (for whatever reason) user runs the Visual Studio for Mac installer as
root using `sudo`, ownership of the `~/.config/xbuild` directory as well as the
configuration file created inside it will be set to root:root. This is due to a
couple of facts:

    * `sudo` (without parameters) switches the user but doesn't reset the
      environment variables. In effect the `$HOME` variable will point to home
      directory of the user running `sudo`, not root
    * Mono's `Environment.GetFolderPath` uses `$HOME` to find the location of
      the user's home directory.

The outcome of the above is that if Xamarin.Android.Tools create the directory
and the config file from scratch, their ownership is going to be incorrect and
that causes the Visual Studio for Mac installer to fail.

In order to fix this issue we detect the above situation and make sure that the
ownership of the directory/file is reset to the original user's ID. This is done
by calling the `chown` Unix utility because this way we avoid the complexity of
calling a number of Unix system functions (getpwent, stat, chown) and the
ownership is reset in the most compatible way.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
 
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -240,5 +303,37 @@ private XDocument GetUnixConfigFile ()
...
 
+		void FixOwnership (List<string> paths)
+		{
+			if (!need_chown || paths == null || paths.Count == 0)
+				return;
+
+			var stdout = new StringWriter ();
+			var stderr = new StringWriter ();
+			var args = new List <string> {
+				QuoteString (sudo_user)
```
jonpryor: I don't see the need for `QuoteString()`. Why not "quote" the string instead of escaping spaces, e.g.:

```csharp
var args = new List<string> {
    $"\"{sudo_user}\"",
};
```

I feel safer about *actually* quoting strings and not escaping certain characters.

grendello: I made the change in `QuoteString`, to keep quoting in a single place should we need to change it.

PR #29, opened 2018-06-25
Jdk properties are required for external clients to make environment …

tondat:
…decisions and informs environment details

JdkProperties get the jdk properties for a given JDK path.
For the moment we are returning values for vendor and java version.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/JdkProperties.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/JdkProperties.cs
@@ -1,0 +1,92 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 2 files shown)


jonpryor:
Conceptually, what we'd prefer to do is "import" this file:

https://github.com/xamarin/java.interop/blob/master/src/Java.Interop.BootstrapTasks/Java.Interop.BootstrapTasks/JdkInfo.cs

presumably as `Xamarin.Android.Tools.AndroidSdk.JdkInfo` (or `Xamarin.Android.Tools.Jdk.JdkInfo`?), with the major change that it is no longer a `Task` subclass, but instead a normal class:

```csharp
public class JdkInfo {
    public JdkInfo (string jdkRoot);

    public string JavaHome {get;}
    public string Vendor {get;}
    public Version Version {get;}

    public static IEnumerable<JdkInfo> GetInstalledJdks (params string[] jdkRoots);
}
```

(This will require a fair bit of refactoring, but it looks plausible/straightforward.)

PR #40, opened 2018-08-01
[tests] Unit tests for finding NDK location based on $PATH

jonpryor:
Commit 511d580b didn't provide unit tests to ensure that the
Android NDK could be found based on probing `$PATH` for `ndk-stack`.

Update `AndroidSdkInfoTests` to ensure that the NDK to be found based
on `$PATH`, as well as the Android SDK and JDK.

Additionally, JDK-via-$PATH detection hit a few hiccups:

 1. `JdkInfo.GetKnownSystemJdkInfos()` would always check
    `/usr/libexec/java_home` before `$PATH`, meaning it wasn't
    possible to test JDK-via-$PATH!

    Alter the ordering of `JdkInfo.GetKnownSystemJdkInfos()` so that
    we probe `$PATH` before hitting `/usr/libexec/java_home`.

 2. `JdkInfo` requires that the `java.home` property be printed, which
    `JdkInfoTests.CreateFauxJdk()` didn't print, which caused the fake
    JDK to be skipped.  Fix `JdkInfoTests.CreateFauxJdk()` so that the
    resulting `java` is acceptable by `JdkInfo`.

Additionally, for greater `monodroid-config.xml` sanity (see also
a4aad186 and b42c2177), instead of backing up the *global*
`monodroid-config.xml` file, instead allow the location of
`monodroid-config.xml` to be overridden by the unit tests by setting a
value accessible by `AppDomain.GetData()`, a'la a4aad186.

Finally, allow the `logger` parameter to
`JdkInfo.GetKnownSystemJdkInfos()` to be optional; if not provided, it
will print to the Console.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -144,7 +144,7 @@ namespace Xamarin.Android.Tools
 			sdk.SetPreferredAndroidNdkPath(path);
 		}
 
-		static void DefaultConsoleLogger (TraceLevel level, string message)
+		internal static void DefaultConsoleLogger (TraceLevel level, string message)
 		{
 			switch (level) {
 			case TraceLevel.Error:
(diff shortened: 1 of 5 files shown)


jonpryor:
> Is it worth adding some kind of check that the system `monodroid-config.xml`

Given that it now should *never* touch the system `monodroid-config.xml` file, I'm not sure any further validation is required.

Issue #39, opened 2018-08-01
JdkInfo.GetReleaseProperties() breaks if values are missing quotes


bratsche:
Using this OpenJDK release https://dl.xamarin.com/OpenJDK/mac/microsoft-dist-openjdk-1.8.0.9.zip which is linked from here https://docs.microsoft.com/en-us/xamarin/android/get-started/installation/openjdk

The release file looks like this

```
JAVA_VERSION="1.8.0"
OS_NAME="Darwin"
OS_VERSION="11.2"
OS_ARCH="x86_64"
SOURCE=""
BUILD_NUMBER=9
```

When `JdkInfo.GetReleaseProperties()` gets to the last line it breaks because it is expecting every value to be inside quotes.  Changing the last line to `BUILD_NUMBER="9"` causes the error to go away.

This line gets added by a Jenkins script http://xamarin-jenkins/view/OpenJDK/job/openjdk-mac/configure but we should probably fix our parser since there are already releases out there.

PR #41, opened 2018-08-01
Don't require quotes around `release` values

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android-tools/issues/39

The `release` file contained within the
[Microsoft OpenJDK Distribution Preview][0] contains an entry which
broke the `release` parser within `JdkInfo`, an entry which doesn't
contain double quotes around the value:

	BUILD_NUMBER=9

This causes an exception to be thrown when `JdkInfo` attempts to parse
the `release` file:

	System.ArgumentOutOfRangeException : Length cannot be less than zero.
	Parameter name: length
	  at System.String.Substring (System.Int32 startIndex, System.Int32 length) [0x0004a] in <7e59c4ad7b424f9eab0db1b78ec1804f>:0
	  at Xamarin.Android.Tools.JdkInfo.GetReleaseProperties () [0x0006b] in <3975ab851981476abded14492926b261>:0
	  at Xamarin.Android.Tools.JdkInfo..ctor (System.String homePath) [0x00159] in <3975ab851981476abded14492926b261>:0

Update `JdkInfo.GetReleaseProperties()` so that the double-quotes
around values are *optional* and not required, so that this `release`
file may be processed without throwing an exception.

[0]: https://dl.xamarin.com/OpenJDK/mac/microsoft-dist-openjdk-1.8.0.9.zip

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
@@ -144,21 +144,26 @@ namespace Xamarin.Android.Tools
 
 		ReadOnlyDictionary<string, string> GetReleaseProperties ()
 		{
+			var props       = new Dictionary<string, string> ();
 			var releasePath = Path.Combine (HomePath, "release");
 			if (!File.Exists (releasePath))
-				return new ReadOnlyDictionary<string, string> (new Dictionary<string, string> ());
+				return new ReadOnlyDictionary<string, string>(props);
 
-			var props   = new Dictionary<string, string> ();
 			using (var release = File.OpenText (releasePath)) {
 				string line;
 				while ((line = release.ReadLine ()) != null) {
-					const string PropertyDelim  = "=\"";
+					line            = line.Trim ();
+					const string PropertyDelim  = "=";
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
```diff
@@ -144,21 +144,25 @@ Version GetJavaVersion ()
...
-					const string PropertyDelim  = "=\"";
+					const string PropertyDelim  = "=";
 					int delim = line.IndexOf (PropertyDelim, StringComparison.Ordinal);
 					if (delim < 0) {
 						props [line] = "";
+						continue;
 					}
 					string  key     = line.Substring (0, delim);
-					string  value   = line.Substring (delim + PropertyDelim.Length, line.Length - delim - PropertyDelim.Length - 1);
+					string  value   = line.Substring (delim + PropertyDelim.Length);
```
jonathanpeppers: Should consider any other weirdness that could happen when parsing this text?

Maybe a call to `Trim ()` for the original `key` and `value`? Such as:
```
KEY_A = 1234 
KEY_B = "Some text here"
```

PR #47, opened 2018-08-14
[tests] fix test failures on Windows

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/blob/b0e6c579f773e55419573e210b648d81b78360ef/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs#L237

Just a few minor things to get these working:
- We need `@echo off` on Windows
- `echo` doesn't need quotes, or it echos the quotes
- We don't need to emit `jvm.dll` as `jvm.dll.cmd`
- There is one test checking `PATH`, which doesn't occur on Windows. It doesn't look in `%PATH%`, so we should ignore it on Windows.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
@@ -74,6 +74,9 @@ namespace Xamarin.Android.Tools.Tests
 		[Test]
 		public void Constructor_SetValuesFromPath ()
 		{
+			if (OS.IsWindows)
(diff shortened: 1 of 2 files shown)


jonpryor:
@jonathanpeppers: Given that these tests have likely never executed on Windows before, I think it would be a good idea to get a quick VSTS job going.

jonathanpeppers:
VSTS Windows job [here](https://devdiv.visualstudio.com/DevDiv/DevDiv%20Team/_build?definitionId=9833&_a=summary). This PR is [green](http://build.devdiv.io/1952778), too!

We have failling tests on master, so I can set this up for CI (and building PRs) after this is merged.

PR #49, opened 2018-09-19
Add path to `apkanalyzer` tool

olegoid:
Needed for this story: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/Xamarin%20VS%20for%20Mac/_workitems/edit/624596

For more background info please refer to this PR:
https://github.com/xamarin/androidtools/pull/143

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -160,9 +160,13 @@ namespace Xamarin.Android.Tools
 			if (string.IsNullOrEmpty (dir))
 				return exe;
 
-			foreach (var e in ProcessUtils.ExecutableFiles (exe))
-				if (File.Exists (Path.Combine (dir, e)))
-					return e;
+			foreach (var e in ProcessUtils.ExecutableFiles (exe)) {
+				var toolPath = Path.Combine (dir, e);
+
+				if (File.Exists (toolPath))
+					return toolPath;
+			}
+			
 			return exe;
 		}
 	}
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -164,13 +164,9 @@ namespace Xamarin.Android.Tools
 			if (string.IsNullOrEmpty (dir))
 				return exe;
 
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -160,9 +164,13 @@ static string GetExecutablePath (string dir, string exe)
...
 				return exe;
 
-			foreach (var e in ProcessUtils.ExecutableFiles (exe))
-				if (File.Exists (Path.Combine (dir, e)))
-					return e;
+			foreach (var e in ProcessUtils.ExecutableFiles (exe)) {
+				var toolPath = Path.Combine (dir, e);
+
+				if (File.Exists (toolPath))
+					return toolPath;
```
pjcollins: I'm worried that you're breaking the expected return value of this function, as it will now return a full path rather than just the tool name with the proper extension appended?

Perhaps you can do a `Path.Combine(sdk.AndroidToolsBinPath, sdk.ApkAnalyzer)` further up in md-addins or wherever you need to actually use this tool, instead of relying on `sdk.ApkAnalyzer` to be a fully qualified path, if this makes sense?

jonathanpeppers:
My concern here is what this directory looks like on Windows...

image

It looks like google has a busted install of `apkanalyzer` on Windows? This file here is a unix script!
```
#!/usr/bin/env sh

##############################################################################
##
##  apkanalyzer start up script for UN*X
##
##############################################################################
```

Is this meant to be used on Mac only? I don't see how it would work on Windows currently...

Does this need to invoke java + the underlying `*.jar` file directly?

/cc @jonpryor

jonpryor:
@jonathanpeppers asked:

> Does this need to invoke java + the underlying `*.jar` file directly?

Yes.

Alternatively, we require Windows Subsystem for Linux and invoke `bash.exe apkanalyzer ...` ;-)  
(That's a terrible idea, don't do that.)

dellis1972:
Also note that historically the `.bat` files on windows are unreliable, and generally have problems with paths containing strings. Which is why we try to call the `.jar` directly.

olegoid:
As we discussed on Slack, apkanalyzer.jar requires classpath to run and this classpath is huge:
https://gist.github.com/olegoid/3ce87a05e21d4dc51217903bcd75ce20#file-gistfile1-txt-L67

Do we have any mechanisms in place to collect/construct classpaths?

jonpryor:
> Do we have any mechanisms in place to collect/construct classpaths?

No.  We use `ToolTask`/`System.Diagnostics.Process` to create processes, and use "normal" command-line arguments.

PR #51, opened 2018-11-28
Add MicroBuild reference + Sign target

cadsit:
We are migrating the signing of this repository to occur in CI via MicroBuild. Other projects have this as their dependency and to make sure that they are correctly signed, we need to push this through.

This should have *no* impact on local builds, other than downloading a new public .nupkg.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
@@ -32,6 +32,18 @@
     <ConsolePause>false</ConsolePause>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="MicroBuild.Core">
+      <Version>0.3.0</Version>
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
```diff
@@ -31,6 +31,18 @@
...
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="MicroBuild.Core">
+      <Version>0.3.0</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+  <ItemGroup>
+    <FilesToSign Include="$(OutDir)\Xamarin.Android.Tools.AndroidSdk.dl">
```
dellis1972: missing `.dll` looks like it

cadsit:
@dellis1972 fixed; the clocks on the build machines were out of sync, causing the test to fail to upload (yay computers!). Should be good to merge?

PR #52, opened 2018-12-03
Check if system Java is installed

bratsche:
This checks for /System/Library/Java/JavaVirtualMachines if we're
trying to use the system java. If there is no system Java installed
then calling /usr/bin/java has the unfortunate effect of popping up
a dialog telling the user to install a JDK.

Fixes VSTS #735545

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735545

PR #53, opened 2018-12-11
[d15-9] Check if system Java is installed

bratsche:
This is a 15-9 backport of https://github.com/xamarin/xamarin-android-tools/pull/52

This checks for /System/Library/Java/JavaVirtualMachines if we're
trying to use the system java. If there is no system Java installed
then calling /usr/bin/java has the unfortunate effect of popping up
a dialog telling the user to install a JDK.

Fixes VSTS #735545


bratsche:
This was already cherry-picked into d15-9 here https://github.com/xamarin/xamarin-android-tools/commit/e3001f4392329b52896e5fe7c3b7952b06f1f68b so closing.

PR #55, opened 2018-12-19
Added ndk-bundle directories to ndk search in Windows

joj:
A few versions ago, the ndk started shipping inside the sdk path. In VS
we are now checking this directories, and only using the registry keys
as fallbacks (meaning, if the ndk exists in the sdk it takes priority
over anything else). According to our telemetry, 100% of the users are
using the ndk this way. This commit adds that check so we can have
parity between what Android does and what VS does. Also, it serves as
preparation for VS removing that option (meaning, it will not be
configurable and _always_ work in the way described in this commit).

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -194,6 +194,17 @@ namespace Xamarin.Android.Tools
 
 			Logger (TraceLevel.Info, "Looking for Android NDK...");
 
+			// Check for the "ndk-bundle" directory inside the SDK directories
+			string ndk;
+
+			var sdks = GetAllAvailableAndroidSdks().ToList();
+			sdks.Add(AndroidSdkPath);
+	
+			foreach(var sdk in sdks.Distinct())
+				if (Directory.Exists(ndk = Path.Combine(sdk, "ndk-bundle")))
+					if (ValidateAndroidNdkLocation(ndk))
+						yield return ndk;
+
 			// Check for the key the user gave us in the VS/addin options
 			foreach (var root in roots)
 				if (CheckRegistryKeyForExecutable (root, regKey, MDREG_ANDROID_NDK, wow, ".", NdkStack))
--- a/src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
@@ -65,6 +65,36 @@ namespace Xamarin.Android.Tools.Tests
 		}
 
 		[Test]
+		public void Ndk_PathInSdk()
+		{
+			if (!OS.IsWindows)
+				Assert.Ignore("This only works in Windows");
+
+			CreateSdks(out string root, out string jdk, out string ndk, out string sdk);
+
+			var logs = new StringWriter();
+			Action<TraceLevel, string> logger = (level, message) => {
+				logs.WriteLine($"[{level}] {message}");
+			};
+
+			try
+			{
+				var ndkPath = Path.Combine(sdk, "ndk-bundle");
+				Directory.CreateDirectory(ndkPath);
+				Directory.CreateDirectory(Path.Combine(ndkPath, "toolchains"));
+				File.WriteAllText(Path.Combine(ndkPath, "ndk-stack.cmd"), "");
+
+				var info = new AndroidSdkInfo(logger, androidSdkPath: sdk, androidNdkPath: null, javaSdkPath: jdk);
+				
+				Assert.AreEqual(ndkPath, info.AndroidNdkPath, "AndroidNdkPath not found inside sdk!");
+			}
+			finally
+			{
+				Directory.Delete(root, recursive: true);
+			}
+		}
+
+		[Test]
 		public void Constructor_SetValuesFromPath ()
 		{
 			if (OS.IsWindows)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -194,6 +194,14 @@ protected override IEnumerable<string> GetAllAvailableAndroidNdks ()
...
 			Logger (TraceLevel.Info, "Looking for Android NDK...");
 
+            // Check for the "ndk-bundle" directory inside the SDK directories
+            string ndk;
+            var sdks = GetAllAvailableAndroidSdks();
+            foreach(var sdk in sdks)
+                if (Directory.Exists(ndk = Path.Combine(sdk, "ndk-bundle")))
+                    if (ValidateAndroidNdkLocation(ndk))
+                        yield return ndk;
+
```
jonathanpeppers: Tabs not spaces...

jonathanpeppers:
Maybe there _is_ CI for this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2288631&view=ms.vss-test-web.test-result-details

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
joj: corrected

joj:
The logs on that link are very windows :). I'll add some tests.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -194,6 +194,17 @@ protected override IEnumerable<string> GetAllAvailableAndroidNdks ()
 
 			Logger (TraceLevel.Info, "Looking for Android NDK...");
 
+			// Check for the "ndk-bundle" directory inside the SDK directories
+			string ndk;
+
+			var sdks = GetAllAvailableAndroidSdks().ToList();
```
jonpryor: Should this instead be:

```csharp
var sdks = new HashSet<string>(GetAllAvailableAndroidSdks (), StringComparer.OrdinalIgnoreCase);
sdks.Add (AndroidSdkPath);
```

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -194,6 +194,17 @@ protected override IEnumerable<string> GetAllAvailableAndroidNdks ()
...
+			sdks.Add(AndroidSdkPath);
+	
+			foreach(var sdk in sdks.Distinct())
```
jonpryor: Please prefer braces.

jonpryor:
This should also update the macOS side of things, in `AndroidSdkUnix.cs`.

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
```diff
@@ -64,6 +64,36 @@ public void Constructor_Paths ()
 			}
 		}
 
+		[Test]
+		public void Ndk_PathInSdk()
+		{
+			if (!OS.IsWindows)
+				Assert.Ignore("This only works in Windows");
```
jonpryor: Android Studio, placed the NDK into `$HOME/Library/Android/sdk/ndk-bundle`.

Unix needs to be supported as well.

joj: The thing is I know how Windows is looking for the NDK (in VS) but I'm not sure what Mac and Unix are doing, so I didn't want to write potentially breaking code. Mac and Unix are not modified right now, and they are in separate classes so they really don't need to. This comment is so the test will always work. There are other tests here doing the opposite (excluding Windows), so it may be common usage to do that here?

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
joj: I was actually thinking on making the default sdk separate and then I don't need the ToList. The thing is all the users I could find of this method end up doing a First(). If I move the default sdk out and make it the first option, and if the sdk and ndk exists (which is always, when the ndk exists at all) then I will never hit the available sdks list (or anything else, for that matter). I'll submit a new PR for review with that change.

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
jonpryor: If mac & Unix are that complicated to figure out, we need to simplify things. :-)

There is a macOS PR builders, so it _is_ possible to ensure that things work there.

joj: I don't think the problem is complexity. The thing is that this logic applies to Windows because of what we do in VS, but I think the existing logic will work for mac just fine, and Unix doesn't have an IDE constraint here. The change is in a class that's specialized to Windows, and the tests are very generic (and that's why I'm ignoring it for non-Windows). There are other tests in this same class that are ignoring Windows explicitly for the same reason.

joj: @sgmunn will probably know if the logic may apply to mac, based on what VSMac is expecting.

sgmunn: @bratsche should have more recent knowledge on this than I do right now.

pjcollins:
Project options which require the NDK are currently busted on VS2019 previews, see https://github.com/xamarin/xamarin-android/issues/2652 for more details. I am not sure if macOS considerations are needed to resolve that particular issue at this time, I believe a fix for windows should be prioritized if possible.

PR #56, opened 2019-01-14
Use a public Azure DevOps project to build, test, and package 'Xamarin.Android.Tools.AndroidSdk'

pjcollins:
An [Azure DevOps build pipeline](https://dev.azure.com/xamarin/Xamarin/_build?definitionId=2&_a=summary) has been configured in the public Xamarin project to build and test Xamarin.Android.Tools on both Windows and macOS. The steps and environments used in this new build pipeline are defined in the new `azure-pipelines.yaml` file, and are described below. A build status badge has also been added to the README.md file which links to the latest successful build. This link can be used to quickly fetch the latest NuGet package produced by this build.

* Windows Steps:
    * Build Xamarin.Android.Tools.sln.
    * Run all unit tests matching the pattern `*-Tests.dll`.
    * Create a NuGet package based on Major, Minor, and Patch versions defined in `azure-pipelines.yaml`. The Patch value starts at 100 and is automatically incremented each time an Azure DevOps build completes.
    * Saves the NuGet package as a build artifact. This build artifact can be downloaded directly from the Azure DevOps build info page. Additionally, this artifact can be published to a public NuGet feed by running the corresponding [Azure DevOps release pipeline](https://dev.azure.com/xamarin/Xamarin/_release?definitionId=1&view=mine).

* macOS Steps:
    * Build Xamarin.Android.Tools.sln.
    * Run all unit tests contained in `Xamarin.Android.Tools.AndroidSdk-Tests.dll`.

This PR also fixes a small bug in an existing unit test which was uncovered by the environments used for the Azure DevOps macOS build pipeline. See https://github.com/xamarin/xamarin-android-tools/commit/2151c3d413dd2530a2c7a1da084548362defcd61 for more information about this test fix.

---
Future considerations:
* The existing [Xamarin.Android.Tools Jenkins build](https://jenkins.mono-project.com/view/Xamarin.Android/job/xamarin-android-tools/) should be retired in favor of this new Azure DevOps pipeline.
* The existing private Azure DevOps build can also be retired.


src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
```diff
@@ -77,6 +77,7 @@ public void Constructor_SetValuesFromPath ()
 				Console.WriteLine ($"[{level}] {message}");
 			};
 			var oldPath = Environment.GetEnvironmentVariable ("PATH");
+			var oldJavaHome = Environment.GetEnvironmentVariable("JAVA_HOME");
```
jonpryor: Code convention: spaces ` ` before parenthesis `(`.

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
```diff
@@ -85,6 +86,8 @@ public void Constructor_SetValuesFromPath ()
 				};
 				paths.AddRange (oldPath.Split (new[]{Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries));
 				Environment.SetEnvironmentVariable ("PATH", string.Join (Path.PathSeparator.ToString (), paths));
+				if (!string.IsNullOrEmpty(oldJavaHome))
```
jonpryor: Code convention: space ` ` before parenthesis `(`.

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
```diff
@@ -94,6 +97,8 @@ public void Constructor_SetValuesFromPath ()
 			}
 			finally {
 				Environment.SetEnvironmentVariable ("PATH", oldPath);
+				if (!string.IsNullOrEmpty(oldJavaHome))
```
jonpryor: Code convention: spaces ` ` before parenthesis `(`.

src/Xamarin.Android.Tools.AndroidSdk/Tests/AndroidSdkInfoTests.cs
jonpryor: Braces should also (generally) be preferred.

azure-pipelines.yaml
```diff
@@ -0,0 +1,61 @@
+name: Xamarin.Android.Tools $(Rev:r)
+variables:
+  major: 1
+  minor: 0
+  patch: $[counter('versioncounter', 100)]
```
jonpryor: What is "patch", and how does it related to "build" and "revision"? https://docs.microsoft.com/en-us/dotnet/api/system.version.-ctor?view=netframework-4.7.2#System_Version__ctor_System_Int32_System_Int32_System_Int32_System_Int32_

*Does it* related to build & revision?  Or is it some NuGet-specific construct?

I'm also not sure what to make of the semantic "NuGet packages are the build number + 100"; presumably we'd wind up with an overlap at *some* point, right?  Unless we "regularly" change the major & minor values more frequently than once every 100 commits?

Assuming we *did* want to "'regularly' change the major & minor values," presumably this means the version numbers should match the Xamarin.Android release values, meaning [the version should be](https://github.com/xamarin/xamarin-android/blob/434d2d888633b11023dfee4cc7cc44ba88666b4e/Configuration.props#L15) v9.2?

I'm really not sure how to sanely handle the version numbers here. :-/

azure-pipelines.yaml
```diff
@@ -0,0 +1,61 @@
...
+  steps:
+  - task: NuGetCommand@2
+    displayName: 'NuGet restore'
+    inputs:
+      restoreSolution: Xamarin.Android.Tools.sln
+  - task: MSBuild@1
+    displayName: 'Build solution Xamarin.Android.Tools.sln'
+    inputs:
+      solution: Xamarin.Android.Tools.sln
+  - task: NuGetCommand@2
```
jonpryor: Instead of this block, should we instead add the `NUnit3TestAdapter` to a `@(PackageReference)` within `Xamarin.Android.Tools.AndroidSdk.csproj` so that it's restored alongside all the other NuGet dependencies, via the `nuget restore Xamarin.Android.Tools.sln` step?

azure-pipelines.yaml
pjcollins: I'm also not sure of the best approach with respect to versioning. The use of "patch" as a variable name here was only done with [Semantic Versioning](https://semver.org/) in mind, and would theoretically map to the `Build` property of `System.Version`.

As for version overlap, I believe the "patch" value will increment ~indefinitely using this approach, so we won't have to worry about changing the major/minor versions for the sole purpose of preventing version overlap.

As for keeping the version in sync with Xamarin.Android releases I don't feel that this should be a requirement, and I am not sure how desirable it is. Perhaps we'd be better served by only updating the major/minor versions alongside code changes in this repository, depending on the scope of those changes?

azure-pipelines.yaml
pjcollins: I agree, though I think it would make sense to add it to the test project (`Xamarin.Android.Tools.AndroidSdk-Tests.csproj`) instead.

jonpryor: Whichever. :-)

pjcollins:
@jonpryor the suggested versioning changes previously discussed (and corresponding README.md updates) have been added in the latest commits.

src/Xamarin.Android.Tools.AndroidSdk/Properties/AssemblyInfo.cs
```diff
@@ -12,4 +12,5 @@
 
 [assembly: AssemblyVersion ("1.0.0.0")]
 [assembly: AssemblyFileVersion ("1.0.0.0")]
+[assembly: AssemblyInformationalVersion ("1.0.99.0")]
```
jonpryor: Should this version differ from `AssemblyVersion`, above?

Relatedly -- which increases complexity immensely, so probably *shouldn't* be done but *could* be considered -- do we want the assembly version values to match the nuget version values?  The only way to do this would be to make `AssemblyInfo.cs` a generated file, so that it can contain the appropriate `$commitsSinceVersionChange` value.

src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.nuspec
```diff
@@ -0,0 +1,16 @@
...
+    <id>$id$</id>
+    <version>$version$</version>
+    <title>$title$</title>
+    <authors>$author$</authors>
+    <owners>$author$</owners>
+    <licenseUrl>https://github.com/xamarin/xamarin-android-tools/blob/master/LICENSE</licenseUrl>
+    <projectUrl>https://github.com/xamarin/xamarin-android-tools</projectUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Xamarin tools for interacting with the Android SDK.</description>
+    <copyright>Copyright 2019</copyright>
```
jonpryor: Presumably this should be a range, at minimum 2016-2019.  (The first commit in this repo is from 2016.)

src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.nuspec
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>$id$</id>
+    <version>$version$</version>
+    <title>$title$</title>
+    <authors>$author$</authors>
```
jonpryor: How does this work/where does `$author$` come from?  I don't see any other relevant substring matches for `author` in this PR.

Ditto the other elements within this `.nuspec` file.

src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.nuspec
jonpryor: For that matter, could this be generated somehow?  `git rev-list --max-parents=0 HEAD` or `git log --reverse` can get the first commit, which can provide the initial year.

pjcollins: I think we should update this to use the `$copyright$` token, which will automatically use the `AssemblyCopyright` value in `AssemblyInfo.cs` Then we'll update `AssemblyInfo.cs` as needed in the future.

More details: https://docs.microsoft.com/en-us/nuget/reference/nuspec

src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.nuspec
pjcollins: These tokens are placeholders for values contained in `AssemblyInfo.cs`. `$author$` maps to `AssemblyCompany`.

jonpryor:
The [**macOS+xbuild PR Build** failed](https://jenkins.mono-project.com/job/xamarin-android-tools-pr-builder/79/):

```
ERROR: Step ‘Publish xUnit test result report’ failed: Clock on this slave is out of sync with the master, and therefore 
I cant figure out what test results are new and what are old.
Please keep the slave clock in sync with the master.
```

I've restarted the build.

jonpryor:
@pjcollins: I asked RE last night to "kick the machine," and they say they've fixed the timestamp issue, so hopefully that won't happen again soon.  (Fingers crossed?)

PR #57, opened 2019-01-25
Made KnownVersions public

joj:
KnownVersions should be the only place we modify when new versions are
published. AndroidVersions right now provides information on what's
installed and what's supported, but not what exists. For Android
Manifest we also need to know what exists because Minimum Android
Version and Target may be anything (regardless of what the compile
version is). This is related to:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/777180


dellis1972:
@joj need to get @jonpryor to confirm he's happy with this.

joj:
@monojenkins backport to d16-0

PR #62, opened 2019-02-06
[Xamarin.Android.Tools.AndroidSdk] Add VS 2017 v15.7 NDK fallback path

brendanzagaeski:
Fixes: https://github.com/xamarin/xamarin-android/issues/2694

Before Visual Studio 2017 version 15.7, the Visual Studio Installer
installed the default Android NDK for the **Mobile development with
C++** workload into:

    [CommonApplicationData]\Microsoft\AndroidNDK64

Starting with Visual Studio 2017 version 15.7, that location changed to:

    [SharedInstallDrive]\Microsoft\AndroidNDK64

Visual Studio 2019 Preview uses that same new location.

This commit adds the new install location to the list of hardcoded NDK
search paths.  It uses the same approach as the **Mobile development
with C++** workload MSBuild targets to get the value of
`[SharedInstallDrive]` based on the `SharedInstallationPath` registry
value.

I tested this change locally with Xamarin.Android apps configured to use
the **Bundle assemblies into native code** and **AOT Compilation
(Experimental)** features.  I verified that the apps failed to build
without this change and built successfully with it.  The builds
correctly located the default Visual Studio 2017 Android NDK
installation even when the `MDREG_ANDROID_SDK` registry value was empty.

I also looked at adding an NUnit test, but that would be tricky because
the test would ideally need to set a registry value, and the
`RegistryEx` class is not visible from the tests.  On top of that, the
`RegistryEx.SetValueString()` method can't write the necessary value in
the `LocalMachine` registry hive without elevated user permissions.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -221,10 +221,19 @@ namespace Xamarin.Android.Tools
 			var vs_default = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.CommonApplicationData), "Microsoft", "AndroidNDK");
 			var vs_default32bit = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.CommonApplicationData), "Microsoft", "AndroidNDK32");
 			var vs_2017_default = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.CommonApplicationData), "Microsoft", "AndroidNDK64");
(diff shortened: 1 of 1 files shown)


jonpryor:
Upon deliberation with a larger team of people, we don't think it's a good idea to support using the `C:\Microsoft\AndroidNDK*` directories because *we* don't control them -- the C++ team does -- and if those directories get updated to a version we don't support -- e.g. NDK r19 is currently not supported by our *stable* releases (xamarin-android master is needed), and NDK r19 will *break* our tooling (directory locations changed) -- then we'll "differently" break.

Consequently, we believe it is safer if we require that NDK that the Visual Studio SDK Manager installs.

PR #65, opened 2019-02-14
[Xamarin.Android.Tools.AndroidSdk][Windows] Allow preferred Android SDK path be empty.

DmitriyKirakosyan:
Use Android SDK path from Windows registry as preferred even if the path doesn't exist in system. The idea is to make Android SDK Installer able to install SDK tools if it wasn't installed within the Xamarin extension, as it is now optional in Willow.

Task #718086

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -65,7 +65,7 @@ namespace Xamarin.Android.Tools
 			androidNdkPath  = androidNdkPath ?? PreferedAndroidNdkPath;
 			javaSdkPath     = javaSdkPath ?? PreferedJavaSdkPath;
 
-			AndroidSdkPath  = ValidateAndroidSdkLocation (androidSdkPath) ? androidSdkPath : AllAndroidSdks.FirstOrDefault ();
+			AndroidSdkPath  = androidSdkPath ?? AllAndroidSdks.FirstOrDefault ();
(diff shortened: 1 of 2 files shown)


dellis1972:
for some reason, this makes me a bit nervous... 
So if I understand this correctly, if a path is provided we will use it, regardless of if it exists or not, and even if it doesn't contain any sdk files...

 @jonpryor ?

DmitriyKirakosyan:
@dellis1972 Correct. The idea is to make SDK Manager work with an empty sdk path. Since Willow is allowing user to skip sdk installation, so that they can install it later, using SDK Manager.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -65,7 +65,7 @@ public virtual void Initialize (string androidSdkPath = null, string androidNdkP
 			androidNdkPath  = androidNdkPath ?? PreferedAndroidNdkPath;
 			javaSdkPath     = javaSdkPath ?? PreferedJavaSdkPath;
 
-			AndroidSdkPath  = ValidateAndroidSdkLocation (androidSdkPath) ? androidSdkPath : AllAndroidSdks.FirstOrDefault ();
+			AndroidSdkPath  = androidSdkPath ?? AllAndroidSdks.FirstOrDefault ();
```
alanmcgovern: Does this line need to change? The net effect of this looks like we just removed validation of the supplied path

DmitriyKirakosyan: Yeah, the idea of the PR is to remove the validation of `PreferredAndroidSdkPath`. Otherwise sdk-manager wouldn't know where to install android sdk if it is absent.

PR #69, opened 2019-06-04
Code that runs in VS needs to target either net46 or net472 in dev16+

kzu:
This makes it fully compatible with NS2. Otherwise, an SDK targeting pack needs to be installed and 
a whole bunch of facade assemblies are included as part of the build which aren't needed and can 
cause issues when running inside the IDE.


kzu:
@kdubau could you help me get this green on mac? thanks!

kzu:
@monokenkins backport to d16-2

PR #71, opened 2019-06-06
[905583]Fix to catch unauthorized exception when config file is created while setting sdkpath

GouriKumari:
Partially Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/905583, Need change in installer code, once this is fixed in xamarin-android-tools.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/905583, VS for Mac installer throws an unauthorized exception, while creating ~/.config files to set sdk path. This PR catches the unhandled exception. 
VS for Mac installer needs a return value to check if the SetPreferredAndroidSDKPath was successful, I changed the  return value of "SetPreferredAndroidSDKPath" to bool to get a return value. I am not sure the impact of this change in other parts of IDE. Let me know if there are other ways to solve this issue. 

A fix for this issue is required, since there are a couple of reports of this unauthorized exception while running installer.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -159,9 +159,9 @@ namespace Xamarin.Android.Tools
 		{
 			logger  = logger ?? DefaultConsoleLogger;
 
-			var sdk = CreateSdk (logger);
-            bool setSDKPath = sdk.SetPreferredAndroidSdkPath(path);
-            return setSDKPath;
+                        var sdk = CreateSdk(logger); 
+                        bool setSDKPath = sdk.SetPreferredAndroidSdkPath(path);
+                        return setSDKPath;
 		}
 
 		public static void SetPreferredJavaSdkPath (string path, Action<TraceLevel, string> logger = null)
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -160,8 +160,7 @@ namespace Xamarin.Android.Tools
 			logger  = logger ?? DefaultConsoleLogger;
 
                         var sdk = CreateSdk(logger); 
-                        bool setSDKPath = sdk.SetPreferredAndroidSdkPath(path);
-                        return setSDKPath;
+                        return sdk.SetPreferredAndroidSdkPath(path);
 		}
 
 		public static void SetPreferredJavaSdkPath (string path, Action<TraceLevel, string> logger = null)
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -155,12 +155,13 @@ namespace Xamarin.Android.Tools
 			}
 		}
 
-		public static void SetPreferredAndroidSdkPath (string path, Action<TraceLevel, string> logger = null)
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -155,12 +155,13 @@ internal static void DefaultConsoleLogger (TraceLevel level, string message)
...
+		public static bool SetPreferredAndroidSdkPath (string path, Action<TraceLevel, string> logger = null)
 		{
 			logger  = logger ?? DefaultConsoleLogger;
 
-			var sdk = CreateSdk (logger);
-			sdk.SetPreferredAndroidSdkPath (path);
+                        var sdk = CreateSdk(logger); 
+                        bool setSDKPath = sdk.SetPreferredAndroidSdkPath(path);
+                        return setSDKPath;
 		}
```
rodrmoya: here, no need for the temp var, just `return sdk.SetPreferredAndroidSdkPath(path);` should be enough

src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -155,12 +155,12 @@ internal static void DefaultConsoleLogger (TraceLevel level, string message)
 			}
 		}
 
-		public static void SetPreferredAndroidSdkPath (string path, Action<TraceLevel, string> logger = null)
+		public static bool SetPreferredAndroidSdkPath (string path, Action<TraceLevel, string> logger = null)
```
jonpryor: This is "public"-ish API, used in numerous places, and it might not be easy to keep all such places synchronized with each other.  This should thus be treated as an ABI break and avoided if at all possible.

Additionally, I fail to see why it would be a "bad" thing for `AndroidSdkInfo.SetPreferredAndroidSdkPath()` to throw an exception when it can't save the specified value.  Why can't the *caller* catch the exception?

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -178,22 +179,30 @@ public override void SetPreferredAndroidNdkPath (string path)
...
+                                    Directory.CreateDirectory(dir);
+                                }
+                                catch (Exception ex)
+                                {
+                                    logger(TraceLevel.Error, $"Unable to create directory {dir}, user do not have the right permissions, {ex.Message}.");
+                                    return false;
+                                }
+                                    AddToList(dir);
+                        }   
+        	   	AddToList (cfg);
```
jonpryor: This file uses *tabs*, not spaces, for indentation.  All the indentation and associated formatting is wrong.

src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
GouriKumari: This issue is caused because user do not have the right permissions to create the `$HOME/.config/xbuild/monodroid-config.xml` and it is caused by https://github.com/xamarin/xamarin-android-tools/pull/71/commits/4dca19a818ff7931240e2165ebb8002fab52031c#diff-c96432cf25d2723c79c03d426742f345L192.  The issue is coming from xamarin-android-tools which is a submodule for installer.  My fix is to handle this issue properly in android-tools, so that installer can handle it  with an error slide and the user knows that installer do not have the right permissions to create $HOME/.config/xbuild/monodroid-config.xml. This is not hiding the error but a way to show user that  there is a permission issue and installer cannot continue Xamarin.Android installation.

GouriKumari:
Closing this PR, this issue will be handled within the installer code.

PR #72, opened 2019-06-07
Get OpenJDK from well-known paths, not registry

kdubau:
Reason: http://work.azdo.io/735565


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -163,6 +162,37 @@ private static IEnumerable<string> GetOpenJdkPaths ()
 			}
 		}
 
+		/// <summary>
+		/// Locate OpenJDK installations by well known path.
+		/// </summary>
+		/// <returns>List of valid OpenJDK paths in version descending order.</returns>
+		private static IEnumerable<string> GetKnownOpenJdkPaths ()
```
kdubau: I did the same as in XamarinVS, https://github.com/xamarin/XamarinVS/blob/82fd693d01115cd38ccdf1d8cdc28b4a0ee65dda/src/Core/VisualStudio.Android/Services/Setup/JavaSdkSetup.cs#L294

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -163,6 +162,37 @@ private static IEnumerable<string> GetOpenJdkPaths ()
...
+		{
+			string JdkFolderNamePattern = "microsoft_dist_openjdk_";
+
+			var paths = new List<Tuple<string, Version>> ();
+			var rootPaths = new List<string>
+			{
+				Path.Combine (Environment.ExpandEnvironmentVariables ("%ProgramW6432%"), "Android", "Jdk"),
```
jonpryor: On my machine, this is `%ProgramW6432%\Android\jdk`, not `...\Jdk`.

kdubau: I don't think case matters on Windows, tested in Workbooks:

image

kdubau: Apparently it _can_ matter if someone is crazy enough to enable such on Windows. Updated anyway.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -50,13 +51,10 @@ public AndroidSdkWindows (Action<TraceLevel, string> logger)
...
 		}
+
 		public override string PreferedJavaSdkPath {
 			get {
-				var wow = RegistryEx.Wow64.Key32;
-				var regKey = GetMDRegistryKey ();
-				if (CheckRegistryKeyForExecutable (RegistryEx.CurrentUser, regKey, MDREG_JAVA_SDK, wow, "bin", JarSigner))
-					return RegistryEx.GetValueString (RegistryEx.CurrentUser, regKey, MDREG_JAVA_SDK, wow);
-				return null;
+				return GetKnownOpenJdkPaths ().FirstOrDefault ();
```
jonpryor: This is where things Go Bad™: the intended semantics is that if you call [`AndroidSdkInfo. SetPreferredJavaSdkPath()`](https://github.com/xamarin/xamarin-android-tools/blob/53ffdaee59e2651a79dec96ad628c5fea851eecf/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs#L166-L172), then the default/preferred Java SDK path will be what was specified:

```
AndroidSdkInfo.SetPreferredJavaSdkPath ("/my/new/JDK/location");
var sdk = new AndroidSdkInfo ();
Console.WriteLine(sdk.JavaSdkPath);
// prints "/my/new/JDK/location"
```

That semantic works because `AndroidSdkWindows.SetPreferredJavaSdkPath()` and `AndroidSdkWindows.PreferedJavaSdkPath` were consistent with each other.

This PR breaks that semantic, and turns `AndroidSdkWindows.SetPreferredJavaSdkPath()` into a no-op, effectively, because the default (preferred) Java SDK may not be what was ever provided to `AndroidSdkWindows.SetPreferredJavaSdkPath()` (unless through sheer happenstance the default value is passed to `AndroidSdkWindows.SetPreferredJavaSdkPath()`...).

Hence my 2019-Mar-26 comment on that issue:

> The way that VS sets a custom path is by setting a registry key.
> 
> If we don't check that registry key *first*, then the *custom path (effectively) does not exist* and it cannot be overridden by the user.

Which returns us to the originally posed (and not yet answered to my satisfaction): *what are we trying to solve*?

Is it *reading* from the registry which is bad?  Or is it just writing?  Both?

Should Windows *copy* Unix and instead use a plain text file to store this information instead of the Registry?  Would that be better?

What about migration and compatibility?  Visual Studio 2019 (Windows) has **Tools** > **Options...** > **Xamarin /  Android Settings**, which allows setting the **Java Development Kit Location** value.  This value in turn (eventually) [uses][0] `AndroidSdkWindows.SetPreferredJavaSdkPath()`.

Which means this change *breaks the IDE* semantics.

[0]: https://github.com/xamarin/XamarinVS/blob/2271dd6b4b69ff844737a280176d960452d6c94b/src/Core/VisualStudio.Android/Services/AndroidToolsSdk.cs#L376

kdubau: Good points, sorry I missed `SetPreferredJavaSdkPath` ...

I will revert so the change doesn't affect this, but at least we'll be able to find based on well-known paths.

PR #73, opened 2019-06-27
Avoids NRE trying to get the list of available NDK if the AndroidSdk path is null

tondat:
- fixes bug 938504: AndroidSdk initialization fails with a System.ArgumentNullException when no AndroidSdk is detected on the user machine
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/938504

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -229,7 +229,8 @@ namespace Xamarin.Android.Tools
 			string ndk;
 
 			var sdks = GetAllAvailableAndroidSdks().ToList();
-			sdks.Add(AndroidSdkPath);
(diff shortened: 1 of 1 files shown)


jonpryor:
https://xamarinhq.slack.com/archives/C58CYJTMY/p1561656927244700

The `ArgumentNullException` is annoying, granted, but if we "fix" it we'll just emit an `InvalidOperationException` instead, from the `AndroidSdkInfo` constructor (as the Android SDK path will be null).

I thus don't see why we *need* this PR.  It might be "nice to have" -- [`InvalidOperationException` stating that the Android SDK couldn't be found](https://github.com/xamarin/xamarin-android-tools/blob/cb41333efe0089581ac3dff8137ec9281a5bbfdf/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs#L22) instead of an `ArgumentNullException` from `Path.Combine()` -- but that's not *necessary*.

joj:
I think the more detailed error _is_ useful. Agreed we don't need to rush it for 16.3, but I would like to see this in master and released whenever.

PR #76, opened 2019-08-20
Support side by side NDK installations in SDK

pjcollins:
The NDK can now be installed by the sdkmanager tool in a side by side
fashion inside the Android SDK. Currently only the `ndk;20.0.5594570`
package exists, and it installs into `$(AndroidSdk)/ndk/20.0.5594570`.

We should re-use the logic used to look up and sort `build-tools` folders
to locate these new NDK installation locations. Additionally, we should
check both the `ndk-bundle` and new `ndk/$(VERSION)` paths on macOS if
we are otherwise unable to locate an NDK installation there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,5 @@ src/*.dll
 bin
 **/obj
 packages/
-TestResult-*.xml
+TestResult*.xml
 .vs/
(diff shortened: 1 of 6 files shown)


pjcollins:
We'll need to consider https://github.com/xamarin/xamarin-android-tools/pull/73 alongside these changes.

pjcollins:
I'm going to close this as the SDK Manager now allows installation of any recent NDK version into the default `ndk-bundle` path, and as a result the desire to support additional detection for versions in `ndk/$(Version)` seems greatly reduced.

PR #78, opened 2019-12-19
Update to sdk-style csproj and netstandard2.0.

jpobst:
- Updates `Xamarin.Android.Tools.AndroidSdk` to use sdk-style projects and target `netstandard2.0` and `net461`.
- Update `Xamarin.Android.Tools.AndroidSdk-Tests` to use sdk-style projects
- Moves `Xamarin.Android.Tools.AndroidSdk-Tests` into a `tests` directory to avoid sdk-style globbing.
- Uses MSBuild properties for NuGet packages instead of .nuspec.
- Updates `MicroBuild.Core` to `Microsoft.VisualStudioEng.MicroBuild.Core`.
  - Not sure what this does but the page said the old one was deprecated and pointed to the new one which supports `netstandard2.0`. ([link](https://www.nuget.org/packages/MicroBuild.Core/))

Original package result:
original

This PR package result:
pr


pjcollins:
I think we'll want to kick off a https://github.com/xamarin/xamarin-android build pointing at this branch for additional testing. This may have some impact on installer creation and projects with a ProjectReference to `Xamarin.Android.Tools.AndroidSdk.csproj`?

jpobst:
This looks good from a `xamarin-android` perspective: 
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3356690&view=results.

kzu:
> Updates MicroBuild.Core to Microsoft.VisualStudioEng.MicroBuild.Core.

This package just provides `build` targets, so it's TF-agnostic. We're using it in SDK-style projects and it just works. It seems to mostly be just a package rename.

PR #79, opened 2020-01-14
[AndroidSdkWindows] Guard against exception checking registry.

jpobst:
In https://developercommunity.visualstudio.com/content/problem/883179/ilegal-characters-in-path-after-fresh-install.html, the user is hitting an exception when checking for an executable from paths found in their registry.

If we hit an error we should just return `false` as there obviously isn't a valid install located there.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -303,19 +303,23 @@ namespace Xamarin.Android.Tools
 		#region Helper Methods
 		private static bool CheckRegistryKeyForExecutable (UIntPtr key, string subkey, string valueName, RegistryEx.Wow64 wow64, string subdir, string exe)
 		{
-			string key_name = string.Format (@"{0}\{1}\{2}", key == RegistryEx.CurrentUser ? "HKCU" : "HKLM", subkey, valueName);
+			try {
+				string key_name = string.Format (@"{0}\{1}\{2}", key == RegistryEx.CurrentUser ? "HKCU" : "HKLM", subkey, valueName);
 
-			var path = NullIfEmpty (RegistryEx.GetValueString (key, subkey, valueName, wow64));
+				var path = NullIfEmpty (RegistryEx.GetValueString (key, subkey, valueName, wow64));
 
-			if (path == null) {
+				if (path == null) {
+					return false;
+				}
+
+				if (!ProcessUtils.FindExecutablesInDirectory (Path.Combine (path, subdir), exe).Any ()) {
+					return false;
+				}
+
+				return true;
+			} catch (Exception) {
 				return false;
 			}
-
-			if (!ProcessUtils.FindExecutablesInDirectory (Path.Combine (path, subdir), exe).Any ()) {
-				return false;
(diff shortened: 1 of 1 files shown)


jonpryor:
In-progress commit message:

```
Fixes: https://developercommunity.visualstudio.com/content/problem/883179/ilegal-characters-in-path-after-fresh-install.html

Somehow, a registry key contains invalid path characters.
This causes an `ArgumentException` from
`AndroidSdkWindows.CheckRegistryKeyForExecutable()`:

	System.ArgumentException: Illegal characters in path.
	   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
	   at System.IO.Path.Combine(String path1, String path2)
	   at Xamarin.Android.Tools.AndroidSdkWindows.CheckRegistryKeyForExecutable(UIntPtr key, String subkey, String valueName, Wow64 wow64, String subdir, String exe)
	   at Xamarin.Android.Tools.AndroidSdkWindows.<GetAllAvailableAndroidSdks>d__37.MoveNext()
	   at System.Linq.Enumerable.<DistinctIterator>d__64`1.MoveNext()
	   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
	   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
	   at Xamarin.Android.Tools.AndroidSdkBase.get_AllAndroidSdks()
	   at Xamarin.Android.Tools.AndroidSdkBase.Initialize(String androidSdkPath, String androidNdkPath, String javaSdkPath)
	   at Xamarin.Android.Tools.AndroidSdkWindows.Initialize(String androidSdkPath, String androidNdkPath, String javaSdkPath)
	   at Xamarin.Android.Tools.AndroidSdkInfo..ctor(Action`2 logger, String androidSdkPath, String androidNdkPath, String javaSdkPath)

Handle the `ArgumentException` by wrapping the body of
`CheckRegistryKeyForExecutable()` within a `try/catch` block and
catch all exceptions.  If an exception is thrown, there is no
executable of interest.
```

jonpryor:
Two questions:

 1. Should the `try`/`catch` block just wrap the `Path.Combine()` call?  That's the only place in the linked-to bug that is throwing; ideally nothing else would throw (we hope?).
 2. Can we/should we log anything as part of the `catch (Exception)` block?  This *can* be done by making `CheckRegistryKeyForExecutable()` an instance method, then using the `Logger` property.

jpobst:
It feels like this was super-unimportant.  The registry is just one of many places we check, and if we can't find it there we move on to the others.  However the one thing it shouldn't do is crash for any reason, because it completely blocks a user from being able to do anything, even if other detection methods would have succeeded.

brendanzagaeski:
*Draft release notes*

Here is the Xamarin.Android release note I plan to include for this pull request:

```
### Issues fixed

#### Application and library build and deployment

  * [Developer Community](https://developercommunity.visualstudio.com/content/problem/883179/ilegal-characters-in-path-after-fresh-install.html):
    *System.ArgumentException: Illegal characters in path* could prevent
    successful automatic detection of the Android SDK location during builds in
    some cases if an `AndroidSdkDirectory` registry value was set for
    Xamarin.Android that contained unexpected characters.
```

To suggest any edits to this, feel free to reply in this PR. Thanks!

PR #83, opened 2020-04-24
Add support for new 'cmdline-tools'

DmitriyKirakosyan:
Google replaced 'tools' with 'cmdline-tools'
Android sdk now supports multiple 'cmdline-tools' versions installed SxS

For the sake of simplicity, using latest installed version for now.

Since new 'cmdline-tools' doesnt contain Emulator executable, replaced the searching path with `Path.Combine (AndroidSdkPath, "emulator")`

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1109288

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -185,5 +185,18 @@ namespace Xamarin.Android.Tools
 			var sdk = CreateSdk (logger);
 			sdk.SetPreferredJavaSdkPath (latestJdk.HomePath);
 		}
+
+		public static string GetPreferredAndroidToolsPath(string androidSdkPath)
+		{
+			var toolsDir = Path.Combine (androidSdkPath, "cmdline-tools");
+			if (!Directory.Exists (toolsDir))
+				return toolsDir;
+
+			if (Directory.Exists (Path.Combine (toolsDir, "latest")))
+				return Path.Combine (toolsDir, "latest");
+
+			var latestToolsDir = Directory.GetDirectories (toolsDir).Max ();
+			return Path.Combine (toolsDir, latestToolsDir);
+		}
 	}
 }
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -52,11 +52,7 @@ namespace Xamarin.Android.Tools
 				foreach (var d in preview)
 					yield return d;
 
-				var sorted = from p in Directory.EnumerateDirectories (buildTools)
-					let version = TryParseVersion (Path.GetFileName (p))
-						where version != null
-					orderby version descending
-					select p;
+				var sorted = SortedSubdirectoriesByVersion (buildTools);
 
 				foreach (var d in sorted)
 					yield return d;
@@ -64,6 +60,15 @@ namespace Xamarin.Android.Tools
 			var ptPath  = Path.Combine (AndroidSdkPath, "platform-tools");
 			if (Directory.Exists (ptPath))
 				yield return ptPath;
+		}
+
+		static IEnumerable<string> SortedSubdirectoriesByVersion (string dir)
+		{
+			return from p in Directory.EnumerateDirectories (dir)
+					let version = TryParseVersion (Path.GetFileName (p))
+						where version != null
+					orderby version descending
+					select p;
 		}
 
 		static Version TryParseVersion (string v)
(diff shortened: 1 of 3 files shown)


JonDouglas:
@dellis1972 For context, we want to support the `tools` directory only if there is no `cmdline-tools` directory for backwards compatibility. We'll check if `cmdline-tools` is on disk first & use that, but if it's not fallback to `tools` since that folder will no longer exist.

src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -185,5 +185,18 @@ public static void DetectAndSetPreferredJavaSdkPathToLatest (Action<TraceLevel,
 			var sdk = CreateSdk (logger);
 			sdk.SetPreferredJavaSdkPath (latestJdk.HomePath);
 		}
+
+		public static string GetPreferredAndroidToolsPath(string androidSdkPath)
+		{
+			var toolsDir = Path.Combine (androidSdkPath, "cmdline-tools");
```
dellis1972, line 191: @JonDouglas  so this method needs to also check for the `tools` if `cmdline-tools` does not exist. Currently if it doesn't exist it will return `cmdline-tools` as per line 193

JonDouglas: @DmitriyKirakosyan ^

DmitriyKirakosyan: @dellis1972 Sounds absolutely reasonable, but I was thinking if supporting old 'tools' could potentially cause any additional issues. If we fallback to it, then Android SDK detection and validation would succeed. But do we really want an SDK using an obsolete component to be valid?

DmitriyKirakosyan: Perhaps avoiding breaking users setup is more important, I agree.
I'll add a fallback.

src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -185,5 +185,24 @@ public static void DetectAndSetPreferredJavaSdkPathToLatest (Action<TraceLevel,
...
+				if (Directory.Exists (oldToolsDir))
+					return oldToolsDir;
+
+				return Path.Combine(toolsDir, "latest");
+			}
+
+			if (Directory.Exists (Path.Combine (toolsDir, "latest")))
+				return Path.Combine (toolsDir, "latest");
+
+			var latestToolsDir = Directory.GetDirectories (toolsDir).Max ();
```
jonpryor, line 204: This isn't quite semantically correct, because `Max()` sorts *lexicographically*, not ordinally:

```csharp
Console.WriteLine (new[]{"1.0", "2.0", "11.0"}.Max ());
// Prints "2.0", because '2' > '1'
```

Fortunately `latest` *always* sorts as the "highest" value…

DmitriyKirakosyan: @jonathanpeppers that is correct, will fix it.

src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -185,5 +190,25 @@ public static void DetectAndSetPreferredJavaSdkPathToLatest (Action<TraceLevel,
 			var sdk = CreateSdk (logger);
 			sdk.SetPreferredJavaSdkPath (latestJdk.HomePath);
 		}
+
+		public static string GetPreferredAndroidToolsPath(string androidSdkPath)
```
jonpryor, line 194: Hopefully the last comment on this PR, and apologies for not seeing this earlier:

Must this method be `static`?  *Most* other methods on this class are *instance* methods, around the `AndroidSdkInfo.AndroidSdkPath` instance property.

I think a better method more "in keeping" with the current design of the class would be:

```csharp
public IEnumerable<string> GetCommandLineToolsPaths (string preferredCommandLineToolsVersion)
{
	if (!string.IsNullOrEmpty (preferredCommandLineToolsVersion)) {
		var preferredDir = Path.Combine (AndroidSdkPath, "cmdline-tools", preferredCommandLineToolsVersion);
		if (Directory.Exists (preferredDir))
			return new[] { preferredDir }.Concat (GetCommandLineToolsPaths (null).Where (p => p != preferredDir));
	}
	return GetCommandLineToolsPaths ();
}

public IEnumerable<string> GetCommandLineToolsPaths ()
{
	var cmdlineToolsDir = Path.Combine (AndroidSdkPath, "cmdline-tools");
	if (Directory.Exists (cmdlineToolsDir)) {
		var latestDir = Path.Combine (cmdlineToolsDir, "latest");
		if (Directory.Exists (latestDir))
			yield return latestDir;
		foreach (var d in SortedSubdirectoriesByVersion (cmdlineToolsDir)) {
			var version = Path.GetFileName (d);
			if (version == "latest")
				continue;
			yield return d;
		}
	}
	var toolsDir = Path.Combine (AndroidSdkPath, "tools");
	if (Directory.Exists (toolsDir)) {
		yield return toolsDir;
	}
}
```

jonpryor:
Draft commit message subject

```
[Xamarin.Android.Tools.AndroidSdk] Add support for cmdline-tools (#83)
```

Body:

```markdown
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1109288
Context: https://dl.google.com/android/repository/repository2-1.xml

Google has deprecated the `tools` Android SDK package, replacing it
with the `cmdline-tools` package, which introduces a version
directory component.

Old and busted:

	$(AndroidSdkDirectory)/tools/bin/sdkmanager

New hotness:

	$(AndroidSdkDirectory)/cmdline-tools/latest/bin/sdkmanager

Of particular interest is that `latest` is a *literal value*.  There
is also a `cmdline-tools;1.0` package which creates a
`cmdline-tools/1.0` directory.

Add a new `AndroidSdkInfo.GetCommandLineToolsPaths()` method which
returns the "command-line tools paths", ordered by version and (non-)
obsolescence.  For example, given the directory structure:

  * `$(AndroidSdkDirectory)/tools/bin/sdkmanager`
  * `$(AndroidSdkDirectory)/cmdline-tools/1.0/bin/sdkmanager`
  * `$(AndroidSdkDirectory)/cmdline-tools/latest/bin/sdkmanager`

Then `AndroidSdkInfo.GetCommandLineToolsPaths()` will return, 
in this order:

  * `$(AndroidSdkDirectory)/cmdline-tools/latest`
  * `$(AndroidSdkDirectory)/cmdline-tools/1.0`
  * `$(AndroidSdkDirectory)/tools`

The `latest` version is always preferred, if present, followed by any
actually versioned cmdline-tools directories, followed by the `tools`
directory, if it exists.

Note that "prefixes" are returned.  All utilities are within a nested
`bin` directory, so if you want e.g. the latest `sdkmanager` util,
you would want to do:

	var info = new AndroidSdkInfo (path);
	var latestSdkManager = Path.Combine (
	        info.GetCommandLineToolsPaths ().First (),
	        "bin",
	        "sdkmanager");
```

PR #86, opened 2020-06-03
[Xamarin.Android.Tools.AndroidSdk] Prefer JAVA_HOME

jonpryor:
Context: https://github.com/xamarin/xamarin-android/pull/4567

`JdkInfo.GetKnownSystemJdkInfos()` returns a list of "known JDK
locations", and the order of the returned paths was, basically,
"locations that Xamarin/Microsoft controls come first."

`AndroidSdkWindows.GetJdkInfos()` reads the Windows registry and is
"controlled by" Visual Studio (unless someone is editing the Registry
by hand…).  If not on Windows, `GetConfiguredJdks()` reads
`monodroid-config.xml` (managed by Visual Studio for Mac); failing
that we probe some "well known Microsoft-controlled" directory
locations…

…and only failing *that* do we try the `$JAVA_HOME` environment
variable, the ["de-facto" way][0] to tell software where Java is
installed, and if `$JAVA_HOME` isn't set we further fallback to
checking directories in `$PATH` and other mechanisms.

The problem with this approach is that it isn't overridable, which is
a usefully important feature if you want to *test new JDK versions*,
as is the case in xamarin/xamarin-android#4567.  The "obvious" way to
"try out" a new JDK would be to export the `JAVA_HOME` environment
variable to the location of the JDK to use, but *that won't work*
because `JdkInfo.GetKnownSystemJdkInfos()` *explicitly prefers*
locations that aren't easily controllable in a CI environment.

Given that *existing convention* is for JDK installs to set the
`JAVA_HOME` environment variable -- and thus `JAVA_HOME` may very
well refer to a JDK which Xamarin.Android doesn't support -- we are
leery of making `JAVA_HOME` the "primary" override.

Instead, add support for a new `JI_JAVA_HOME` environment variable
which, if set, is the *preferred* JDK to use within Xamarin.Android
(unless otherwise overridden by e.g. `$(JavaSdkDirectory)`).
This will allow CI to export the `JAVA_HOME` environment variable,
allowing it to be preferred over others.

Additionally, remove some `JAVA_HOME` "duplication" between `JdkInfo`
and `AndroidSdkWindows`, so that things are easier to reason about.

[0]: https://docs.oracle.com/cd/E19182-01/821-0917/inst_jdk_javahome_t/index.html

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
@@ -281,10 +281,11 @@ namespace Xamarin.Android.Tools
 		{
 			logger  = logger ?? AndroidSdkInfo.DefaultConsoleLogger;
 
-			return GetJavaHomeEnvironmentJdks (logger)
+			return GetEnvironmentVariableJdks ("JI_JAVA_HOME", logger)
 				.Concat (GetWindowsJdks (logger))
 				.Concat (GetConfiguredJdks (logger))
 				.Concat (GetMacOSMicrosoftJdks (logger))
+				.Concat (GetEnvironmentVariableJdks ("JAVA_HOME", logger))
 				.Concat (GetPathEnvironmentJdks (logger))
 				.Concat (GetLibexecJdks (logger))
 				.Concat (GetJavaAlternativesJdks (logger))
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
It looks like this is also in two places?
https://github.com/xamarin/xamarin-android-tools/blob/967c278b22fb3636a7ece8189800baaf8d2b6b94/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs#L138-L142

jpobst:
My concern is that we don't know who all is setting `JAVA_HOME`.  What if some installer out there is setting it to a Java distro or version we don't support?  Today we would do the "right thing" for the user and use a supported Java.  With this change they will have to go change their `JAVA_HOME` (and perhaps that other program they use needed it set the way it was).

It feels like it might be safer to introduce a `XA_JAVA_HOME` so the user can override our preferred JDK if they want, but it won't be overridden by another Java installer?

jonpryor:
> Can we remove the `[Ignore]` from this test now?

We *could*, if we used `JAVA_HOME`, but it looks like @jpobst  and @pjcollins would prefer using `XA_JAVA_HOME` instead.

I should certainly add a test for whatever env var is used.

jonpryor:
> It looks like this is also in two places?

Hilarious.  I should remove that. :-)

jonpryor:
The "extra" `GetEnvironmentJdkPaths()` invocation within `AndroidSdkWindows` should now be removed, and the `AndroidSdkInfo` unit test will now attempt to override `JI_JAVA_HOME`.

All comments should thus be resolved (I hope).

PR #89, opened 2020-06-29
Add API 29 and 30 to KnownVersions

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/4873

Although this is related to https://github.com/xamarin/xamarin-android/pull/4873, it is not the full fix. We probably should list the missing versions here anyway, though.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
@@ -176,6 +176,12 @@ namespace Xamarin.Android.Tools
 			new AndroidVersion (28, "9.0",   "Pie") {
 				AlternateIds = new[]{ "P" },
 			},
+			new AndroidVersion (29, "10.0",  "Android10") {
+				AlternateIds = new[]{ "Q" },
+			},
+			new AndroidVersion (30, "11.0") {
+				AlternateIds = new[]{ "R" },
+			},
 		};
 	}
 


jonpryor:
For "sanity" purposes, I would prefer for starters that we *not* update `KnownVersions`, and instead try the patch suggested on https://github.com/xamarin/xamarin-android/pull/4873#issuecomment-651331337:

```diff
diff --git a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
index 1297e51..d2d3b5f 100644
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -98,11 +98,15 @@ namespace Xamarin.Android.Tools
 
 		public string? TryGetPlatformDirectoryFromApiLevel (string idOrApiLevel, AndroidVersions versions)
 		{
+			string? dir = GetPlatformDirectoryFromId (idOrApiLevel);
+			if (Directory.Exists (dir))
+				return dir;
+
 			var id  = versions.GetIdFromApiLevel (idOrApiLevel);
 			if (id == null)
 				return null;
 
-			string? dir = GetPlatformDirectoryFromId (id);
+			dir = GetPlatformDirectoryFromId (id);
 
 			if (Directory.Exists (dir))
 				return dir;
```

and try to see how far we can get things to work w/o having `KnownVersions` data.

jonathanpeppers:
@jonpryor I think the problem happens even sooner:

https://github.com/xamarin/xamarin-android-tools/blob/3974fc38c0f25f943b5d3bf0a4e174532a2a60ee/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs#L105-L119

If 29 came in here, these would both return `null`. And so `null` is passed into `TryGetPlatformDirectoryFromApiLevel` as well.

jonathanpeppers:
@jonpryor I can confirm that the actual problem here is fixed in: https://github.com/xamarin/net6-samples/pull/25

This is using: https://github.com/xamarin/xamarin-android/commit/5d1ac0a5dfa12e92fc45b384237e929576a418e4

Should we merge this, or should I close it?

src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
```diff
@@ -173,6 +173,10 @@ static bool MatchesId (AndroidVersion version, string id)
 			new AndroidVersion (28, "9.0",   "Pie") {
 				AlternateIds = new[]{ "P" },
 			},
+			new AndroidVersion (29, "10.0",  "Android X") {
```
jonpryor, line 176: "Android X" was the codeName for API-29?!

I don't remember that codename…

src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
```diff
@@ -173,6 +173,10 @@ static bool MatchesId (AndroidVersion version, string id)
...
+				AlternateIds = new[]{ "Q" },
+			},
+			new AndroidVersion (30, "11.0",   "R"),
```
jonpryor, line 179: This should probably just omit the `codeName` parameter, and instead include `R` in the `AlternateIds`:

```csharp
new AndroidVersion (30, "11.0") {
    AlternateIds = new[]{ "R" },
},
```

src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
jonathanpeppers: I guess it is `Android10`?

https://source.android.com/setup/start/build-numbers

PR #91, opened 2020-06-30
Fix sort ordering for ndk-bundle, add macOS support

jonathanpeppers:
Context: https://build.azdo.io/xamarin/public/21545

An environment change on Azure DevOps seems to cause the test failure
on Windows:

    Ndk_PathInSdk
    Expected: "C:\Users\VssAdministrator\AppData\Local\Temp\tmpAE78.tmp\sdk\..."
    But was: "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"

It appears that `xamarin-android-tools` is preferring an `ndk-bundle
found on the system instead of the value passed in for
`AndroidSdkPath`.

To make matters worse, use of `Distinct()` returns an unordered
sequence, and so the sort ordering is unknown:

https://docs.microsoft.com/dotnet/api/system.linq.enumerable.distinct

To fix this, let's not use `ToList()` or add `AndroidSdkPath` to the
list at all. We can make an explicit check for it before the `foreach`
loop. Inside the loop we should skip `AndroidSdkPath`.

I also added the same code for the macOS/linux side in
`AndroidSdkUnix`.

I updated the `Ndk_PathInSdk` test so it will run on all platforms.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -10,20 +10,19 @@ namespace Xamarin.Android.Tools
 	abstract class AndroidSdkBase
 	{
 		string[]? allAndroidSdks;
-		string[]? allAndroidNdks;
 
 		public string[] AllAndroidSdks {
 			get {
-				if (allAndroidSdks == null)
-					allAndroidSdks = GetAllAvailableAndroidSdks ().Distinct ().ToArray ();
+				if (allAndroidSdks == null) {
+					var dirs = new List<string?> ();
+					dirs.Add (AndroidSdkPath);
+					dirs.AddRange (GetAllAvailableAndroidSdks ());
+					allAndroidSdks = dirs.Where (d => ValidateAndroidSdkLocation (d))
+						.Select (d => d!)
+						.Distinct ()
+						.ToArray ();
+				}
 				return allAndroidSdks;
-			}
-		}
-		public string[] AllAndroidNdks {
-			get {
-				if (allAndroidNdks == null)
-					allAndroidNdks = GetAllAvailableAndroidNdks ().Distinct ().ToArray ();
-				return allAndroidNdks;
 			}
 		}
 
@@ -57,13 +56,10 @@ namespace Xamarin.Android.Tools
 
 		public virtual void Initialize (string? androidSdkPath = null, string? androidNdkPath = null, string? javaSdkPath = null)
 		{
-			androidSdkPath  = androidSdkPath ?? PreferedAndroidSdkPath;
-			androidNdkPath  = androidNdkPath ?? PreferedAndroidNdkPath;
-			javaSdkPath     = javaSdkPath ?? PreferedJavaSdkPath;
+			AndroidSdkPath = GetValidPath (ValidateAndroidSdkLocation,  androidSdkPath, () => PreferedAndroidSdkPath,   () => GetAllAvailableAndroidSdks ());
+			JavaSdkPath    = GetValidPath (ValidateJavaSdkLocation,     javaSdkPath,    () => PreferedJavaSdkPath,      () => GetJavaSdkPaths ());
 
-			AndroidSdkPath  = ValidateAndroidSdkLocation (androidSdkPath) ? androidSdkPath : AllAndroidSdks.FirstOrDefault ();
-			AndroidNdkPath  = ValidateAndroidNdkLocation (androidNdkPath) ? androidNdkPath : AllAndroidNdks.FirstOrDefault ();
-			JavaSdkPath     = ValidateJavaSdkLocation (javaSdkPath) ? javaSdkPath : GetJavaSdkPath ();
+			AndroidNdkPath = GetValidNdkPath (androidNdkPath);
 
 			if (!string.IsNullOrEmpty (JavaSdkPath)) {
 				JavaBinPath = Path.Combine (JavaSdkPath, "bin");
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
Wait this isn't green:
```
1) Failed : Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Constructor_SetValuesFromPath
  AndroidNdkPath not set from $PATH!
  Expected string length 68 but was 44. Strings differ at index 1.
  Expected: "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp11587757...."
  But was:  "/Users/runner/Library/Android/sdk/ndk-bundle"
  ------------^
  at Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Constructor_SetValuesFromPath () [0x00000] in <4e091f45935d4073b64b6415de873d6b>:0 
```

jonpryor:
Related question: https://github.com/xamarin/xamarin-android-tools/blob/3974fc38c0f25f943b5d3bf0a4e174532a2a60ee/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs#L61

What should be the semantics of `AndroidSdkInfo.AndroidNdkPath` relative to `.AndroidSdkPath` when `AndroidSdkInfo` is constructed with `androidSdkPath` set (and valid), but `androidNdkPath` *not* set?

Should `AndroidSdkInfo.AndroidNdkPath` be the *default system-wide* path?  Or should it be `{androidSdkPath}/ndk-bundle`?

Given that the local `androidNdkPath` variable/parameter is set *before* `AllAndroidNdks`/`GetAllAvailableAndroidNdks()` is accessed, it seems that the *former* would be the behavior, even with the current state of the PR.

I *believe* that what is more "reasonable" is the *latter* behavior: that *if* `androidSdkPath` is set, *and* it contains an `ndk-bundle`, then `{androidSdkPath}/ndk-bundle` should be preferred *over* an existing "default" system-wide preferred NDK dir.

In order for that to happen, I think we'd need to *remove* `AndroidSdkInfo.cs`:61.  I think that would be "fine".

*Relatedly*, `AllAndroidNdks`/`GetAllAvailableAndroidNdks()` should return `PreferedAndroidNdkPath` "first" (or "second", after `{androidSdkPath}/ndk-bundle}, should `{androidSdkPath}/ndk-bundle` exist).

jonpryor:
Related: https://github.com/xamarin/xamarin-android-tools/issues/92

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -10,22 +10,21 @@ namespace Xamarin.Android.Tools
...
 			get {
-				if (allAndroidSdks == null)
-					allAndroidSdks = GetAllAvailableAndroidSdks ().Distinct ().ToArray ();
+				if (allAndroidSdks == null) {
+					var dirs = new List<string?> ();
+					dirs.Add (AndroidSdkPath);
+					dirs.AddRange (GetAllAvailableAndroidSdks ());
+					allAndroidSdks = dirs.Where (d => ValidateAndroidSdkLocation (d))
+						.Select (d => d!)
+						.Distinct ()
```
jonathanpeppers, line 22: @jonpryor will `Distinct()` lose the ordering again? or does the ordering not matter here?

jonpryor: Yes, it will lose the ordering again, but this is only from `AndroidSdkInfo.AllAndroidSdks`, which "nobody" should be using, and which *always* had `.Distinct()` semantics, so anybody using that method shouldn't see anything bizarre.

I hope.

joj: Confirmed we're only using GetAllAvailableAndroidSdks (and also every time we use it we're using Distinct too :)

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android-tools/issues/92

Context: https://github.com/xamarin/xamarin-android-tools/pull/90#issuecomment-651375196

The PR builds for #90 encountered an "unrelated test failure" 
in `AndroidSdkInfoTests.Ndk_PathInSdk()` on Windows, because Windows
is non-deterministic: the test asserts that given an Android SDK
directory `androidSdk` which contains the file
`{androidSdk}\ndk-bundle\ndk-stack.cmd`, then this:

	var info = new AndroidSdkInfo (logger: null, androidSdkPath: androidSdk);

will have `info.AndroidNdkPath`==`{androidSdk}\ndk-bundle`.

Instead, this test would occasionally fail on CI:

	Ndk_PathInSdk
	Expected: "C:\Users\VssAdministrator\AppData\Local\Temp\tmpAE78.tmp\sdk\..."
	But was: "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"

Here, the "preferred"/system-wide NDK is being chosen over the
`{androidSdk}\ndk-bundle` directory that the unit test created.

The wrong directory was chosen for two reasons: 

 1. `AndroidSdkBase.Initialize()` would use `PreferedAndroidNdkPath`
    when `androidNdkPath` was null, *first*, before we checked
    `{androidSdk}\ndk-bundle`.

 2. If `PreferedAndroidNdkPath` happened to be null, then
    `AndroidSdkBase.Initialize()` would try to use
    `AllAndroidNdks.FirstOrDefault()` as a default value, also before
    checking `{androidSdk}\ndk-bundle`.  The problem here is that the
    `AllAndrdoidNdks` property uses [`Enumerable.Distinct()`][0], which
    returns an *unordered* list of directories.

That the test ever worked at all is a minor miracle.

Additionally, the support for `{androidSdk}\ndk-bundle` was Windows-
specific; it didn't run on macOS.

Update `AndroidSdkInfo` so that `{androidSdk}/ndk-bundle` is supported
on macOS, and that `{androidSdk}/ndk-bundle` is *preferred* when the
`androidNdkPath` parameter is `null`, *before* checking any other
plausible default locations.

This allows the `AndroidSdkInfoTests.Ndk_PathInSdk()` test to run
everywhere, and work reliably.

[0]: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=netcore-3.1
```

PR #93, opened 2020-07-01
Add a .props file which contains the Android SDK Tools versions.

dellis1972:
This commit is part of an effort to make sure all of our tooling uses
the same android sdk versions. It adds a MSbuild `.props` file
which contains the versions of parts of the Android Sdk. It is
these versions which will be installed/used by default when a
user first installs the product, or if they do not specify
other versions in their `csproj`.

If the new file is changed ALL repos referencing this one
will need to be updated. This includes all repos which pull
in this one via `androidtools`.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
@@ -37,4 +37,9 @@
     </FilesToSign>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="Xamarin.Android.Tools.Versions.props">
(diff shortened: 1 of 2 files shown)


dellis1972:
failure does not look related.

```
X Ndk_PathInSdk [150ms]
  Error Message:
  Expected: "C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\tmp4A25.tmp\\sdk\\..."
  But was:  "C:\\Program Files (x86)\\Android\\android-sdk\\ndk-bundle"
```

jonpryor:
This should also add a `$(AndroidCommandLineToolsVersion)` value, a'la:

https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in#L12

PR #94, opened 2020-07-01
[build] fail macOS build if tests fail

jonathanpeppers:
Context: https://docs.microsoft.com/azure/devops/pipelines/tasks/test/publish-test-results#yaml-snippet
Context: https://dev.azure.com/xamarin/public/_build/results?buildId=21614&view=ms.vss-test-web.build-test-results-tab

Currently, if a test fails on macOS the build will still be green.

We need to set `failTaskOnFailedTests` on the `PublishTestResults` step.

I also filled out `testRunTitle`, to fix the display name shown on
test results.

### DIFF
--- a/azure-pipelines.yaml
+++ b/azure-pipelines.yaml
@@ -39,6 +39,7 @@ jobs:
     displayName: 'Run Tests'
     inputs:
       testAssemblyVer2: 'bin\TestDebug\*-Tests.dll'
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
The build here failed appropriately now: https://dev.azure.com/xamarin/public/_build/results?buildId=21660&view=ms.vss-test-web.build-test-results-tab

PR #95, opened 2020-07-08
Generate constants for the RequiredComponentVersions.

dellis1972:
This commit adds support for reading this new props file and auto generating the C# code
for the properties on the RequiredComponentVersions class. This removes the requirement that
the file be on disk or that xamarin.android is installed.


joj:
You guys were actually right and I was wrong, this is better placed in sdk installer

PR #96, opened 2020-08-25
Localization 

moljac:


workflow/Localization.md
```diff
@@ -0,0 +1,73 @@
+# Localization
```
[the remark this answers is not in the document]
moljac: Well I saw that file in the xamarin-android commit, but did not know what should I do, so I added it here, thinking doc would not hurt.
Removed.

src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
```diff
@@ -29,6 +34,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="XliffTasks" Version="1.0.0-beta.20420.1" />
```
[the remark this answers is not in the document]
moljac: So. I guess no changes necessary here. Or am I wrong?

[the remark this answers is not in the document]
brendanzagaeski: Correct. No changes needed on the version number

But now that I'm looking at this again, I realized one tiny detail that could be added:

```diff
-<PackageReference Include="XliffTasks" Version="1.0.0-beta.20420.1" />
+<PackageReference Include="XliffTasks" Version="1.0.0-beta.20420.1" PrivateAssets="all" />
```

In other words, the package can be set to `PrivateAssets="all"` to match [dotnet/arcade](https://github.com/dotnet/arcade/blob/91470b0b14ba016c1fb78211b12775287c17b34e/src/Microsoft.DotNet.Arcade.Sdk/tools/Localization.targets#L26) and [xamarin-android](https://github.com/xamarin/xamarin-android/blob/ac4039d309a36467bf982df0a8f30941b5fb9642/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj#L56).

src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
```diff
@@ -45,7 +46,7 @@ public JdkInfo (string homePath)
 			if (homePath == null)
 				throw new ArgumentNullException (nameof (homePath));
 			if (!Directory.Exists (homePath))
-				throw new ArgumentException ("Not a directory", nameof (homePath));
+				throw new ArgumentException ("XamarinAndroidTools_XAT0008", nameof (homePath));
```
jonpryor, line 49: Shouldn't this use `Resources.ResourceManager.GetString("XamarinAndroidTools_XAT0008")`

moljac: Yes. Sorry

moljac: Fixed.

moljac: Added one more which was missed.

src/Xamarin.Android.Tools.AndroidSdk/Properties/Resources.resx
```diff
@@ -0,0 +1,206 @@
...
+  
+  <data name="XamarinAndroidTools_XAT0012" xml:space="preserve">
+    <value>GetMacOSMicrosoftJdkPaths jdks override! {0}</value>
+    <comment>GetMacOSMicrosoftJdkPaths jdks override! {0}</comment>
+  </data>
+  
+  <data name="XamarinAndroidTools_XAT0013" xml:space="preserve">
+    <value>The directory `{0}`, via locator `{1}`, is not a valid JDK directory: {2}</value>
+    <comment>The directory `{0}`, via locator `{1}`, is not a valid JDK directory: {2}</comment>
+  </data>
```
brendanzagaeski, line 184: The `comment` field won't need to contain a copy of the `value`. It can instead be used to provide additional context about the string as needed.

While we're on the topic of this message, I'll suggest adjusting the wording to make it more actionable and to align it more closely with some other messages from xamarin-android and androidtools:

```xml
<data name="XamarinAndroidTools_XAT0013" xml:space="preserve">
  <value>An exception occurred while validating the Java SDK installation in '{0}' that was found while searching the paths in '{1}'. Ensure that the Android section of the Visual Studio options has a valid Java SDK directory configured. To use a custom SDK path for a command line build, set the 'JavaSdkDirectory' MSBuild property to the custom path. Exception: {2}</value>
  <comment>{0} - The path of the invalid installation
{1} - The search path
{2} - The exception message of the associated exception</comment>
</data>
```

brendanzagaeski (changes requested):
Thanks again for the work on this! I'll leave a few comments in the individual files and add a few broader discussion points directly in this comment.

1. The updated `logger()` call in `JdkInfo` is heading in exactly the right direction. It looks like there are a few more `logger()` calls that also need to be updated to be localizable:

   ```sh
   $ git grep -n "[lL]ogger (TraceLevel\.\(Err\|Warn\)"
   src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs:321:				logger (TraceLevel.Warning, $"Not a valid JDK directory: `{path}`; via locator: {locator}");
   src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs:253:					logger (TraceLevel.Error, "Could not load monodroid configuration file");
   src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs:295:				Logger (TraceLevel.Warning, $"Failed to change ownership of filesystem object(s)");
   src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs:122:					logger (TraceLevel.Warning, $"Not a valid JDK directory: `{path}`; via category: {locator}");
   ```

   - The Xamarin.Android build tasks only ever hit the calls in `JdkInfo.cs`, `AndroidSdkUnix.cs:253`, and `AndroidSdkWindows.cs` indirectly via `JdkInfo.GetKnownSystemJdkInfos()`, so all three share the same logging mechanism at the point where they actually get logged in MSBuild.
   - `AndroidSdkUnix.cs:295` is within `FixOwnership()`, which does not appear to be used by any Xamarin.Android build tasks, so it can be skipped for now.

2. A small complication is that these `logger()` messages are not yet associated with any warning codes in the MSBuild tasks. `JdkInfo.GetKnownSystemJdkInfos()` is hit in just one place in the MSBuild tasks:

   [`<ResolveJdkJvmPath/>`](https://github.com/xamarin/xamarin-android/blob/40644d91d8486123afc837dc416dbb928a8172fc/src/Xamarin.Android.Build.Tasks/Tasks/ResolveJdkJvmPath.cs#L58):

   ```cs
   info = JdkInfo.GetKnownSystemJdkInfos (this.CreateTaskLogger ()).FirstOrDefault ();
   ```

   [`CreateTaskLogger()`](https://github.com/xamarin/xamarin-android/blob/40644d91d8486123afc837dc416dbb928a8172fc/src/Xamarin.Android.Build.Tasks/Utilities/MSBuildExtensions.cs#L188-L204) does not currently assign any codes to these messages:

   ```cs
   public static Action<TraceLevel, string> CreateTaskLogger (this AsyncTask task)
   {
   	Action<TraceLevel, string> logger = (level, value) => {
   		switch (level) {
   		case TraceLevel.Error:
   			task.LogError (value);
   			break;
   		case TraceLevel.Warning:
   			task.LogWarning (value);
   			break;
   		default:
   			task.LogDebugMessage (value);
   			break;
   		}
   	};
   	return logger;
   }
   ```

   For comparison, in androidtools, the small number of logging statements that do end up as MSBuild errors and warnings are all surfaced via the [error and warning handlers in `<ResolveXamarinAndroidTools/>`](https://github.com/xamarin/monodroid/blob/1562e30689e60b884a0b229a233bb6d7bf230a08/tools/msbuild/Tasks/ResolveXamarinAndroidTools.cs#L135-L143), which set XA5300 as the code for all the messages.

   I think `<ResolveJdkJvmPath/>` and `CreateTaskLogger()` can be modified slightly to assign a single code to all of the errors and warnings that come from the xamarin-android-tools `logger()` calls. That way, xamarin-android-tools won't itself need to worry about incorporating message codes into the messages, at least for now.

   If this sounds OK, then the only change needed in this PR is to make the other `logger()` calls from (1) localizable. The xamarin-android side can take care of adding the codes.

Side note: The only place the output of the `<ResolveJdkJvmPath/>` Xamarin.Android MSBuild task is currently used is for the `_GenerateJniMarshalMethods` target.

jonpryor:
@brendanzagaeski's [comment](https://github.com/xamarin/xamarin-android-tools/pull/96#pullrequestreview-493846717) raises a pre-existing-yet-unanswered question around error code management: How should they be managed?

The interaction between xamarin/Java.Interop & xamarin/xamarin-android has the "same" problem/scenario: `Xamarin.Android.Build.Tasks.dll` (xamarin-android) uses `Xamarin.Android.Tools.Bytecode.dll` (Java.Interop). If the latter throws an exception, how should we manage the error codes?

Similar-but-different, *should* the "leaf methods" *know* about error codes in the first place?  `Xamarin.Android.Tools.Bytecode.dll` is used by *both* `Xamarin.Android.Build.Tasks.dll` *and* `class-parse.exe`.  Should these *different* contexts share the same error "codes", wherein the code includes a "prefix" such as XA1234?  Should `class-parse` emit a `XA1234` error and not a `CP1001` error?

With this "philosophical thinking" in place, I think the xamarin-android-tools repo *shouldn't* know, care, or otherwise worry about error codes.  I think that should be handled by the *callers*, e.g. the `<ResolveJdkJvmPath/>` task.

However, this poses a different question: different errors should get different codes.  How should this "difference" be communicated?

In the case of xamarin-android-tools, there is no difference; there is only one use of `TraceLevel.Error`:

```bash
% git grep TraceLevel.Error  
src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs:                 case TraceLevel.Error:
src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs:                                    logger (TraceLevel.Error, "Could not load monodroid configuration file");
```

That doesn't answer how things should work in e.g. Java.Interop, but for here-and-now, we don't need to worry about error codes *at all*.

jpobst:
Note that currently this PR adds several NRT warnings to the build: https://dev.azure.com/xamarin/public/_build/results?buildId=27103&view=results.

jonpryor:
Context: https://github.com/xamarin/xamarin-android/commit/0342fe5698b86e21e36c924732ff135b9a87e4af

jonpryor:
@brendanzagaeski: Question: how do we make the generated `Resource.Designer.cs` emit an *internal* type?  `Xamarin.Android.Build.Tasks.dll` has an internal type, but the one in this repo is `public`.

jonpryor:
Commit message for squash-and-merge review:

```markdown
Context: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1009374/
Context: https://github.com/xamarin/xamarin-android/commit/0342fe5698b86e21e36c924732ff135b9a87e4af

Localize error, warning, and exception messages produced by
`Xamarin.Android.Tools.AndroidSdk.dll`.

We will be following the [.NET Resource Localization pattern][0] and
generating satellite assemblies using [`.resx` files][1], in particular
`src/Xamarin.Android.Tools.AndroidSdk/Properties/Resources.resx`.

`Resources.resx` is an XML file, and will contain `/root/data`
elements in which `//data/@name` will start with the
xamarin-android-tools error or warning code, and `//data/value` will
be the error or warning message:

	<root>
	  <data name="XamarinAndroidTools_XAT0001" xml:space="preserve">
	    <value>App manifest does not have 'manifest' root element</value>
	    <comment>The following are literal names and should not be translated: manifest</comment>
	  </data>
	</root>

An optional `//data/comment` element may be provided to describe the
meaning within the `//data/value` element to translators.

When using Visual Studio or Visual Studio for Mac, changes to
`Resrouces.resx` will cause `Resources.Designer.cs` to be updated:

	namespace Xamarin.Android.Tools.AndroidSdk.Properties {
	  internal partial class Resources {
	    internal static string XamarinAndroidTools_XAT0001 {
	      get => ...
	    }
	  }
	}

The `Resources` members should be used to obtain all strings for use
in `logger()` and `throw new …()` calls:

	logger (TraceLevel.Warning, string.Format (Resources.XamarinAndroidTools_XAT0013, path, locator, e.Message));
	throw new ArgumentException (Resources.XamarinAndroidTools_XAT0001, nameof (doc));

When an MSBuild error or warning code is used with more than one
output string, then a semantically meaningful suffix should be used
to distinguish between the two.

Note that this infrastructure does not interoperate with C#6 string
interpolation.  Any error or warning messages currently using C#6
string interpolation will need to use .NET 1.0-style format strings.

Our translation team doesn't work directly with `.resx` files.
Instead, the translation team works with [XLIFF files][2].
`Resources.resx` is converted into a set of
`src/Xamarin.Android.Build.Tasks/Properties/xlf/Resources.*.xlf`
files via `XliffTasks.targets` from the [dotnet/xliff-tasks][3] repo.
The `Resources.*.xlf` files should be automatically updated whenever
`Resources.resx` is updated.


[0]: https://docs.microsoft.com/dotnet/framework/resources/index
[1]: https://docs.microsoft.com/dotnet/framework/resources/creating-resource-files-for-desktop-apps#resources-in-resx-files
[2]: http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html
[3]: https://github.com/dotnet/xliff-tasks
```

brendanzagaeski:
To make the generated members `internal`, adjust `src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj`:

```diff
-<Generator>PublicResXFileCodeGenerator</Generator>
+<Generator>ResXFileCodeGenerator</Generator>
```

azure-pipelines.yaml
```diff
@@ -31,6 +31,13 @@ jobs:
     displayName: 'NuGet Restore'
     inputs:
       restoreSolution: Xamarin.Android.Tools.sln
+      feedsToUse: config
+      nugetConfigPath: NuGet.config
+  - task: MSBuild@1
+    displayName: 'Update Xlf'
+    inputs:
+      solution: 'src\Xamarin.Android.Tools.AndroidSdk\Xamarin.Android.Tools.AndroidSdk.csproj'
+      msbuildArguments: '/t:UpdateXlf'
```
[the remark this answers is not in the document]
moljac: The problem was - it was failing on the CI without those lines.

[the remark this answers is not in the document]
moljac: OK I see JonP already remoevd it

src/Xamarin.Android.Tools.AndroidSdk/AndroidAppManifest.cs
```diff
@@ -33,7 +34,7 @@ public class AndroidAppManifest
 			this.doc = doc;
 			manifest = doc.Root;
 			if (manifest.Name != "manifest")
-				throw new ArgumentException ("App manifest does not have 'manifest' root element", nameof (doc));
+				throw new ArgumentException (Resources.ResourceManager.GetString ("XamarinAndroidTools_XAT0001"), nameof (doc));
```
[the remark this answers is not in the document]
moljac: > with apologies that this will mean discarding some of this nice work you set up.

Why apologies? It is my job. Sorry that I did not know that exceptions are being kept out of the scope

Started reverting, but will leave in `Resources.resx` for now. Unless you want it differently

workflow/Localization.md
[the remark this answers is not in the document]
brendanzagaeski: No problem. Either way is OK, really. Since it's removed now, we can leave it like it is. I'll mark this conversation as resolved. Thanks!

src/Xamarin.Android.Tools.AndroidSdk/Properties/Resources.resx
```diff
@@ -0,0 +1,129 @@
...
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  
+  <data name="InvalidJdkDirectory" xml:space="preserve">
+    <value>The directory `{0}`, via locator `{1}`, is not a valid JDK directory: {2}</value>
+    <comment>The following are literal names and should not be translated: JDK.
+{0} - a directory
+{1} - a "contextual" name for where {0} came from: `Preferred Registry` (Windows Registry), `OpenJDK`, `$JAVA_HOME` (environment variable), etc.
+{2} - the `Exception.Message` from the exception prompting us to show this message.</comment>
+  </data>
```
brendanzagaeski, line 127: No worries if my [previous suggestion](https://github.com/xamarin/xamarin-android-tools/pull/96#discussion_r493099061) for wording didn't look good, but in case it was just missed unintentionally in the hustle and bustle, I'll re-suggest that new wording to make the message more actionable and to align it more closely with some similar messages from xamarin-android and androidtools:

```
<data name="InvalidJdkDirectory" xml:space="preserve">
  <value>An exception occurred while validating the Java SDK installation in '{0}' that was found while searching the paths in '{1}'. Ensure that the Android section of the Visual Studio options has a valid Java SDK directory configured. To use a custom SDK path for a command line build, set the 'JavaSdkDirectory' MSBuild property to the custom path. Exception: {2}</value>
  <comment>{0} - The path of the invalid installation
{1} - The search path
{2} - The exception message of the associated exception</comment>
</data>
```

brendanzagaeski (changes requested):
The two other `logger()` calls [mentioned previously](https://github.com/xamarin/xamarin-android-tools/pull/96#pullrequestreview-493846717) are not yet localizable:

```
src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs:235:                                        logger (TraceLevel.Error, "Could not load monodroid configuration file");
src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs:117:                                     logger (TraceLevel.Warning, $"The directory `{path}`, via locator `{locator}`, is not a valid JDK directory: {e.Message}");
```

I believe it will make sense to include those in this PR since they are surfaced as MSBuild messages in the same way as the `logger()` call in `JdkInfo`. Thanks!

_Wording brainstorming:_

- For:

   ```
   src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs:235:                                        logger (TraceLevel.Error, "Could not load monodroid configuration file");
   ```

   I'll suggest something like:

   ```xml
   <data name="InvalidJdkPathConfigFile" xml:space="preserve">
     <value>An exception occurred while loading the Java SDK path configuration file '{0}'. Ensure that the file is a valid XML file. Exception: {1}</value>
     <comments>{0} - The path of the invalid file
   {1} - The exception message of the associated exception</comments>
   </data>
   ```

- For:

   ```
   src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs:117:                                     logger (TraceLevel.Warning, $"The directory `{path}`, via locator `{locator}`, is not a valid JDK directory: {e.Message}");
   ```

   This one can use the same `Resources.InvalidJdkDirectory` resource as used in `JdkInfo`.

Thanks!

src/Xamarin.Android.Tools.AndroidSdk/Properties/Resources.resx
```diff
@@ -117,12 +117,23 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+
+  <data name="FailedToChangeFileSystemOwnership" xml:space="preserve">
+    <value>Failed to change ownership of filesystem object(s).</value>
```
brendanzagaeski, line 121: `object(s)` is at least sometimes translated as if it were just `objects`. For example, `item(s)` was translated as the plural `elementos` in the following example from [dotnet/msbuild](https://github.com/dotnet/msbuild/blob/ca44138662e3aa90eb9305dd31d906ef02e962cb/src/Tasks/Resources/xlf/Strings.es.xlf#L588-L589):

```xml
<source>MSB3094: "{2}" refers to {0} item(s), and "{3}" refers to {1} item(s). They must have the same number of items.</source>
<target state="translated">MSB3094: "{2}" hace referencia a {0} elementos y "{3}" hace referencia a {1} elementos. Deben tener el mismo número de elementos.</target>
```

Although for some target languages there won't be any way to improve this because the correct grammar requires multiple different wordings for different numbers of items, a small improvement for at least some target languages could be to use the phrase "one or more":

```
Failed to change the ownership of one or more filesystem objects.
```

On a different topic, this message doesn't yet suggest a resolution action, but from what I could find, `FixOwnership()` isn't called by any Xamarin.Android build tasks, so it's not currently expected that Xamarin.Android users will see this message during a build anyway. Considering that, this message could optionally be omitted from localization for now, but it's OK to include it too.

jonpryor: Actually, considering that we "know" that this assembly is used by *an installer*, but we *don't* know if said installer will actually capture & redistribute the satellite assemblies containing the translated strings, it's probably better if we *don't* translate this message.

brendanzagaeski (approved):
Approved! I pushed one more small commit to update the generated files, so those should be all set now. 

It'll probably be good to get the build pipeline to run once more on this PR to be careful before merging, but it looks good locally, so I think that'll build without problems if we can get it to trigger.

brendanzagaeski:
I manually triggered the pipeline for this branch via the Azure Pipelines UI. That seemed to allow it to run correctly, including having `$(RunningOnCi)`=`true` so that the validity of the `.xlf` files was checked as expected. In short, I think this is all set to merge now.

PR #98, opened 2020-10-01
[Xamarin.Android.Tools.AndroidSdk] Prefer API-28 Component Versions

jonpryor:
We were late in attempting to update dependencies, resulting in an
"undesirable" first run experience on macOS+Visual Studio for Mac, in
which creating a new Xamarin.Forms project would elicit a dialog to
the user:

> **Android SDK components needed**
>
> The project … requires the following Android SDK
> components that are not currently installed:
>
> -- Android SDK Platform 28 v6
>
> The download will take a few minutes…
>
> [Cancel] [Download and Install]

…because the default template would be using Xamarin.Android
`$(TargetFrameworkVersion)`=v9.0 (API-28), which wasn't installed.

"Revert" MSBuild properties so that API-28 -related packages are
referenced, which will allow them to be installed on macOS, which will
thus prevent the above dialog from being shown:

  * `$(AndroidSdkBuildToolsVersion)`=28.0.3
  * `$(AndroidSdkPlatformToolsVersion)`=28.0.2
  * `$(AndroidSdkPlatformVersion)`=android-28

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props
@@ -11,10 +11,10 @@
     -->
     <AndroidSdkBuildToolsVersion Condition="'$(AndroidSdkBuildToolsVersion)' == ''">29.0.2</AndroidSdkBuildToolsVersion>
     <AndroidCommandLineToolsVersion Condition=" '$(AndroidCommandLineToolsVersion)' == '' ">1.0</AndroidCommandLineToolsVersion>
-    <AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">29.0.5</AndroidSdkPlatformToolsVersion>
+    <AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">30.0.4</AndroidSdkPlatformToolsVersion>
     <AndroidSdkToolsVersion Condition="'$(AndroidSdkToolsVersion)' == ''">26.1.1</AndroidSdkToolsVersion>
     <AndroidSdkEmulatorVersion Condition="'$(AndroidSdkEmulatorVersion)' == ''"></AndroidSdkEmulatorVersion>
-    <AndroidSdkPlatformVersion Condition="'$(AndroidSdkPlatformVersion)' == ''">android-29</AndroidSdkPlatformVersion>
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props
```diff
@@ -9,12 +9,12 @@
     If this file is changed the submodule for androidtools should be updated, 
     along with any other repo which references androidtools. 
     -->
-    <AndroidSdkBuildToolsVersion Condition="'$(AndroidSdkBuildToolsVersion)' == ''">29.0.2</AndroidSdkBuildToolsVersion>
+    <AndroidSdkBuildToolsVersion Condition="'$(AndroidSdkBuildToolsVersion)' == ''">28.0.3</AndroidSdkBuildToolsVersion>
     <AndroidCommandLineToolsVersion Condition=" '$(AndroidCommandLineToolsVersion)' == '' ">1.0</AndroidCommandLineToolsVersion>
-    <AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">29.0.5</AndroidSdkPlatformToolsVersion>
+    <AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">28.0.2</AndroidSdkPlatformToolsVersion>
```
brendanzagaeski, line 14: I think for the smoothest update experience, the `AndroidSdkBuildToolsVersion` and `AndroidSdkPlatformToolsVersion` can be left on the `29.0.2` and `29.0.5` versions to match d16-7:

https://github.com/xamarin/xamarin-android/blob/d16-7/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in#L18-L19

GouriKumari: $(AndroidSdkPlatformVersion)=android-28,  is the only entry that  changed in release-8.8. We can keep other tools version as such as Brendan mentioned.

jonpryor: Relevant comparison point isn't xamarin-android, but android-sdk-installer: https://github.com/xamarin/android-sdk-installer/blob/d16-7/Xamarin.Installer.AndroidSDK/Xamarin.Installer.AndroidSDK/Common/Constants.cs

…which does specify a version for `RequiredBuildToolsVersion` (29.0.2), but not for the *Platform-tools* version.

@GouriKumari? On your fresh VSMac 8.8 install, what's the output of this command?

```bash
cat ~/Library/Developer/Xamarin/android-sdk-macosx/platform-tools/source.properties
```

GouriKumari (approved):

PR #101, opened 2020-12-15
Add Microsoft.Android.Build.BaseTasks project

pjcollins:
We are hoping to sever the xamarin-android dependency from monodroid.
In order to do this we'll want to share some common MSBuild code across
our two build task assemblies, `Xamarin.Android.Build.Tasks` and
`Xamarin.Android.Build.Debugging.Tasks`.

This common code has been moved from xamarin/xamarin-android into
a new project in this repo named `Microsoft.Android.Build.BaseTasks`.
More utilities can be migrated here in the future if we so desire.

A new test project `Microsoft.Android.Build.BaseTasks-Tests.csproj` has
also been added to bring over the relevant tests from xamarin-android.


src/Microsoft.Android.Build.BaseTasks/AndroidRidAbiHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+			if (runtimeIdentifier.EndsWith ("-arm", StringComparison.OrdinalIgnoreCase)) {
+				return "armeabi-v7a";
+			}
+			if (runtimeIdentifier.EndsWith ("-x86", StringComparison.OrdinalIgnoreCase)) {
+				return "x86";
+			}
+			if (runtimeIdentifier.EndsWith ("-x64", StringComparison.OrdinalIgnoreCase)) {
+				return "x86_64";
+			}
+			return "";
```
jonathanpeppers, line 89: Is it worth "fixing" anything?

I think we should make this check that it starts with `android`, as seen here we could get `linux` native libraries:

https://github.com/xamarin/xamarin-android/pull/5386

src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs
```diff
@@ -0,0 +1,88 @@
...
+		public abstract bool RunTask ();
+	}
+
+	public abstract class AndroidAsyncTask : AsyncTask
+	{
+		/// <summary>
+		/// A helper for non-async overrides of RunTaskAsync, etc.
+		/// </summary>
+		public static readonly System.Threading.Tasks.Task Done =
+			System.Threading.Tasks.Task.FromResult (true);
```
jonathanpeppers, line 33: We can remove this, and use `System.Threading.Tasks.Task.CompletedTask` instead.

I think I added this when I didn't know about `CompletedTask`.

src/Microsoft.Android.Build.BaseTasks/AndroidToolTask.cs
```diff
@@ -0,0 +1,122 @@
...
+		//        if (String.Compare (category, "warning", StringComparison.OrdinalIgnoreCase) == 0) {
+		//                Log.LogWarning (null, null, null, filename, lineNumber, 0, -1,
+		//                        -1, text, null);
+		//        } else if (String.Compare (category, "error", StringComparison.OrdinalIgnoreCase) == 0) {
+		//                Log.LogError (null, null, null, filename, lineNumber, 0, -1,
+		//                        -1, text, null);
+		//        } else {
+		//                Log.LogMessage (importance, singleLine);
+		//        }
+		//}
```
jonathanpeppers, line 69: Maybe go ahead and delete all this commented code.

src/Microsoft.Android.Build.BaseTasks/MSBuildExtensions.cs
```diff
@@ -0,0 +1,255 @@
...
+namespace Microsoft.Android.Build.Tasks
+{
+	public static class MSBuildExtensions
+	{
+		private static bool IsRunningInsideVS {
+			get { 
+				var vside = false;
+				return bool.TryParse(Environment.GetEnvironmentVariable("VSIDE"), out vside) && vside; 
+			}
+		}
```
jonathanpeppers, line 21: Looks like you can delete this one.

src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs
dellis1972: Good call. Also do we want to split these up into their own files since we are moving them?

src/Microsoft.Android.Build.BaseTasks/Microsoft.Android.Build.BaseTasks.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <GenerateResxSource>true</GenerateResxSource>
+    <UpdateXlfOnBuild Condition=" '$(RunningOnCI)' != 'true' ">true</UpdateXlfOnBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.Framework" Version="16.5.0" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="16.5.0" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="16.5.0" IncludeAssets="compile" />
+    <PackageReference Include="Xamarin.Build.AsyncTask" Version="0.3.4" IncludeAssets="compile" />
+    <PackageReference Include="Xamarin.LibZipSharp" Version="1.0.20" IncludeAssets="compile" />
```
pjcollins, line 18: @jonathanpeppers  /  @dellis1972  do you think it makes sense to try to better sync these versions with xamarin-android and monodroid? Maybe we should move part of https://github.com/xamarin/xamarin-android/blob/master/build-tools/scripts/MSBuildReferences.projitems here?

src/Microsoft.Android.Build.BaseTasks/MSBuildExtensions.cs
pjcollins: Removed 👍

src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs
pjcollins: I updated this to use CompletedTask and split up all of the Android Task extensions into their own files.

src/Microsoft.Android.Build.BaseTasks/AndroidRidAbiHelper.cs
pjcollins: I added this check to the first if statement so we will bail early if the RID doesn't start with `android`.

src/Microsoft.Android.Build.BaseTasks/Microsoft.Android.Build.BaseTasks.csproj
jonathanpeppers: I think moving these to their own MSBuild file makes sense. They could be imported from other repos.

The only problem I see is the libZipSharp version. Is it OK to move the version to this repo? libZipSharp bumps would go from xamarin-android-tools then to xamarin-android?

The libZipSharp version is currently here:

https://github.com/xamarin/xamarin-android/blob/087c7f43a1a60a6c92934d0a57c902e381877df3/Configuration.props#L127

pjcollins: Yeah I'm wondering if this will introduce more of a headache than the alternative -- keeping versions in multiple repos in sync...

The biggest issue I see with LibZipSharp is: https://github.com/xamarin/xamarin-android/blob/bea063b939d7d8d5f166c50564665d62ab140417/build-tools/xaprepare/xaprepare/ThirdPartyNotices/LibZipSharp.cs#L16, but I suppose we could update that to read from a file in this repo instead.

I'll move the version properties into a separate file that can be referenced elsewhere for now, and we can decide if we want to use it everywhere later.

dellis1972: We have this file https://github.com/xamarin/xamarin-android-tools/blob/main/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props which is imported in xa 
so perhaps the LibZipSharp versions can go in there?

dellis1972: although its not imported as part of the build only in Xamarin.Android.Common.props.

pjcollins: I think we ship the `Xamarin.Android.Tools.Versions.props` file in our installers as well, so it may be preferable to add a new file for these product build dependency related items?

src/Microsoft.Android.Build.BaseTasks/AndroidAsyncTask.cs
```diff
@@ -0,0 +1,49 @@
...
+		{
+			try {
+				return RunTask ();
+			} catch (Exception ex) {
+				this.LogUnhandledException (TaskPrefix, ex);
+				return false;
+			}
+		}
+
+		public virtual bool RunTask ()
```
jonpryor, line 27: Should this actually be `virtual`, and not *non*-virtual?

src/Microsoft.Android.Build.BaseTasks/AndroidAsyncTask.cs
```diff
@@ -0,0 +1,49 @@
...
+	{
+		/// <summary>
+		/// A helper for non-async overrides of RunTaskAsync, etc.
+		/// </summary>
+		public static readonly System.Threading.Tasks.Task Done =
+			System.Threading.Tasks.Task.CompletedTask;
+
+		public abstract string TaskPrefix { get; }
+
+		public override bool Execute ()
```
jonpryor, line 17: Should this be `override sealed`, so that it can't be further overridden?

src/Microsoft.Android.Build.BaseTasks/AndroidToolTask.cs
```diff
@@ -0,0 +1,24 @@
...
+				return RunTask ();
+			} catch (Exception ex) {
+				Log.LogUnhandledException (TaskPrefix, ex);
+				return false;
+			}
+		}
+
+		// Most ToolTask's do not override Execute and
+		// just expect the base to be called
+		public virtual bool RunTask () => base.Execute ();
```
jonpryor, line 22: I'm not sure I understand how this "works" w/o causing stack overflow if the subclass doesn't "do" anything:

```csharp
class ExampleTask : AndroidToolTask {
    public override string TaskPrefix => "FOO";
}
```

Wouldn't `AndroidToolTask.Execute()` all `AndroidToolTask.RunTask()` call `.Execute()` call `.RunTask()` call…?

"Feels like" this should instead have `RunTask()` be `abstract`, and `Execute()` should be `override sealed`.

src/Microsoft.Android.Build.BaseTasks/XDocumentExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+		static ITaskItem ToTaskItem (XElement element)
+		{
+			var taskItem = new TaskItem (element.Value);
+			foreach (var attribute in element.Attributes ()) {
+				taskItem.SetMetadata (attribute.Name.LocalName, attribute.Value);
+			}
+			return taskItem;
+		}
+
+		public static IEnumerable<XElement> ToXElements (ICollection<ITaskItem> items, string itemName, string[] knownMetadata)
```
jonpryor, line 34: This isn't an extension method; should it be placed into a separate type?  Or made non-`public`?

src/Microsoft.Android.Build.BaseTasks/AndroidAsyncTask.cs
dellis1972: No, because under certain circumstances we still need to override execute. One example is when calling the MSBuild GetRegisterTaskObject, that needs to be done from the UI thread.

src/Microsoft.Android.Build.BaseTasks/AndroidToolTask.cs
jonathanpeppers: This is:
* `AndroidToolTask.Execute()`
* `AndroidToolTask.RunTask()`
* `base.Execute()` or `ToolTask.Execute()` from MSBuild

Hopefully, should not be a stack overflow.

I don't think we can mark anything `sealed` as mentioned: https://github.com/xamarin/xamarin-android-tools/pull/101#discussion_r551219608

src/Microsoft.Android.Build.BaseTasks/XDocumentExtensions.cs
jonathanpeppers: @pjcollins can we change this one to an extension method?

So the usage changes to:
```diff
--XDocumentExtensions.ToXElements (ResolvedResourceDirectories, "ResolvedResourceDirectory", knownMetadata)
++ResolvedResourceDirectories.ToXElements ("ResolvedResourceDirectory", knownMetadata)
```

pjcollins: I moved this to an `ITaskItemExtensions` class. However, since this is not used by monodroid, we could just remove it and put a new `ITaskItemExtensions.cs` in `xamarin/xamarin-android` instead?

jonathanpeppers: If monodroid doesn't use it, we can keep it in xamarin-android.

src/Microsoft.Android.Build.BaseTasks/AndroidAsyncTask.cs
pjcollins: I added a comment for this method.

src/Microsoft.Android.Build.BaseTasks/XDocumentExtensions.cs
pjcollins: I removed it, we can make this change in xamarin-android instead.

src/Microsoft.Android.Build.BaseTasks/AndroidAsyncTask.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Microsoft.Android.Build.Tasks
+{
+	public abstract class AndroidAsyncTask : AsyncTask
+	{
+		/// <summary>
+		/// A helper for non-async overrides of RunTaskAsync, etc.
+		/// </summary>
+		public static readonly System.Threading.Tasks.Task Done =
+			System.Threading.Tasks.Task.CompletedTask;
```
jonathanpeppers, line 13: I think we can just remove this member, and callers can use `System.Threading.Tasks.Task.CompletedTask` directly.

pjcollins: I removed this, it looks like it's only used in one place in xamarin-android and is easy enough to replace.

pjcollins:
This should be ready to go, but we're going to wait to merge this until the week of the 18th or later. There are a lot of connected pieces that should land at the same time, and we want to give d16-9 a little more time before merging all of them as all of these changes will make backporting difficult.

jonathanpeppers (approved):

PR #103, opened 2021-01-20
[NDK] Locate and select only compatible NDK versions

grendello:
Context: https://github.com/xamarin/xamarin-android/issues/5499
Context: https://github.com/xamarin/xamarin-android/issues/5526
Context: https://github.com/android/ndk/issues/1427
Context: https://developer.android.com/studio/command-line/variables#envar

Xamarin.Android is not (yet) compatible with the recently released
Android NDK r22 version.  Azure build images have recently rolled out an
update which includes NDK r22 and, thus, it breaks builds for customers
using any form of Xamarin.Android AOT build.

In attempt to detect broken/incompatible NDK versions as well as select
the "best one", this commit adds code to scan the known NDK locations in
search of the preferred version.  The search is conducted as follows:

  1. If the user selected a preferred NDK location, it is always used.
  2. Locations specified in the `ANDROID_{HOME,SDK_ROOT}` environment
     variables are returned next.
  3. Directories in the `PATH` environment variable are examined to find
     a valid NDK location.
  4. OS-specific known NDK locations are considered.

For each of the returned locations, we now look for the Android SDK
packages containing the NDK.  There are two kinds of such packages:

  * `ndk-bundle` is the older package which allows for installation of
    only one NDK inside the SDK directory
  * `ndk/*` is a newer package which allows for installation of several
    NDK versions in parallel.  Each subdirectory of `ndk` is an `X.Y.Z`
    version number of the NDK.

In each of these directories we look for the `source.properties` file
from which we then extract the NDK version and then we sort thus
discovered NDK instances using their version as the key, in the
descending order.  The latest compatible (currently: less than 22 and
more than 15) version is selected and its path returned to the caller.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -9,6 +9,15 @@ namespace Xamarin.Android.Tools
 {
 	abstract class AndroidSdkBase
 	{
+		// When this changes, update the test: Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Ndk_MultipleNdkVersionsInSdk
+		const int MinimumCompatibleNDKMajorVersion = 16;
+		const int MaximumCompatibleNDKMajorVersion = 21;
+
+		static readonly char[] SourcePropertiesKeyValueSplit = new char[] { '=' };
+
+		// Per https://developer.android.com/studio/command-line/variables#envar
+		protected static readonly string[] AndroidSdkEnvVars = {"ANDROID_HOME", "ANDROID_SDK_ROOT"};
+
 		string[]? allAndroidSdks;
 
 		public string[] AllAndroidSdks {
@@ -97,8 +106,9 @@ namespace Xamarin.Android.Tools
 			if (pathValidator (ctorParam))
 				return ctorParam;
 			foreach (var path in getAllPaths ()) {
-				if (pathValidator (path))
+				if (pathValidator (path)) {
 					return path;
+				}
 			}
 			return null;
 		}
@@ -108,7 +118,7 @@ namespace Xamarin.Android.Tools
 			if (ValidateAndroidNdkLocation (ctorParam))
 				return ctorParam;
 			if (AndroidSdkPath != null) {
-				string bundle = Path.Combine (AndroidSdkPath, "ndk-bundle");
+				string bundle = FindBestNDK (AndroidSdkPath);
 				if (Directory.Exists (bundle) && ValidateAndroidNdkLocation (bundle))
 					return bundle;
 			}
@@ -125,6 +135,18 @@ namespace Xamarin.Android.Tools
 		protected abstract IEnumerable<string> GetAllAvailableAndroidSdks ();
 		protected abstract string GetShortFormPath (string path);
 
+		protected IEnumerable<string> GetSdkFromEnvironmentVariables ()
+		{
+			foreach (string envVar in AndroidSdkEnvVars) {
+				string ev = Environment.GetEnvironmentVariable (envVar);
+				if (String.IsNullOrEmpty (ev)) {
+					continue;
+				}
+
+				yield return ev;
+			}
+		}
+
 		protected virtual IEnumerable<string> GetAllAvailableAndroidNdks ()
 		{
 			// Look in PATH
@@ -139,7 +161,67 @@ namespace Xamarin.Android.Tools
 			foreach (var sdk in GetAllAvailableAndroidSdks ()) {
 				if (sdk == AndroidSdkPath)
 					continue;
-				yield return Path.Combine (sdk, "ndk-bundle");
+				yield return FindBestNDK (sdk);
+			}
+		}
+
+		string FindBestNDK (string androidSdkPath)
+		{
+			var ndkInstances = new SortedDictionary<Version, string> (Comparer<Version>.Create ((Version l, Version r) => r.CompareTo (l)));
+
+			foreach (string ndkPath in Directory.EnumerateDirectories (androidSdkPath, "ndk*", SearchOption.TopDirectoryOnly)) {
+				if (String.Compare ("ndk-bundle", Path.GetFileName (ndkPath), StringComparison.OrdinalIgnoreCase) == 0) {
+					LoadNDKVersion (ndkPath);
+					continue;
+				}
+
+				if (String.Compare ("ndk", Path.GetFileName (ndkPath), StringComparison.OrdinalIgnoreCase) != 0) {
+					continue;
+				}
+
+				foreach (string versionedNdkPath in Directory.EnumerateDirectories (ndkPath, "*", SearchOption.TopDirectoryOnly)) {
+					LoadNDKVersion (versionedNdkPath);
+				}
+			}
+
+			if (ndkInstances.Count == 0) {
+				return String.Empty;
+			}
+
+			var kvp = ndkInstances.First ();
+			Logger (TraceLevel.Verbose, $"Best NDK selected: v{kvp.Key} in {kvp.Value}");
+			return kvp.Value;
+
+			void LoadNDKVersion (string path)
+			{
+				string propsFilePath = Path.Combine (path, "source.properties");
+				if (!File.Exists (propsFilePath)) {
+					Logger (TraceLevel.Verbose, $"Skipping NDK in '{path}': no source.properties, cannot determine version");
+					return;
+				}
+
+				foreach (string line in File.ReadLines (propsFilePath)) {
+					string[] parts = line.Split (SourcePropertiesKeyValueSplit, 2, StringSplitOptions.RemoveEmptyEntries);
+					if (parts.Length != 2) {
+						continue;
+					}
+
+					if (String.Compare ("Pkg.Revision", parts[0].Trim (), StringComparison.Ordinal) != 0) {
+						continue;
+					}
+
+					if (!Version.TryParse (parts[1].Trim (), out Version? ndkVer) || ndkVer == null || ndkInstances.ContainsKey (ndkVer)) {
+						continue;
+					}
+
+					if (ndkVer.Major < MinimumCompatibleNDKMajorVersion || ndkVer.Major > MaximumCompatibleNDKMajorVersion) {
+						Logger (TraceLevel.Verbose, $"Skipping NDK in '{path}': version {ndkVer} is out of the accepted range (major version must be between {MinimumCompatibleNDKMajorVersion} and {MaximumCompatibleNDKMajorVersion}");
+						continue;
+					}
+
+					ndkInstances.Add (ndkVer, path);
+					return;
+				}
 			}
 		}
 
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
@@ -94,6 +94,10 @@ namespace Xamarin.Android.Tools
 			if (!string.IsNullOrEmpty (preferedSdkPath))
 				yield return preferedSdkPath!;
 
+			foreach (string dir in GetSdkFromEnvironmentVariables ()) {
+				yield return dir;
+			}
+
 			// Look in PATH
 			foreach (var adb in ProcessUtils.FindExecutablesInPath (Adb)) {
 				var path = Path.GetDirectoryName (adb);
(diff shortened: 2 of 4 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -9,6 +9,14 @@ namespace Xamarin.Android.Tools
 {
 	abstract class AndroidSdkBase
 	{
+		const int MinimumCompatibleNDKMajorVersion = 16;
+		const int MaximumCompatibleNDKMajorVersion = 21;
+
+		static readonly char[] SourcePropertiesKeyValueSplit = new char[] { '=' };
+
+		// Per https://developer.android.com/studio/command-line/variables#envar
+		protected static readonly string[] AndroidSdkEnvVars = {"ANDROID_HOME", "ANDROID_SDK_ROOT"};
```
jonathanpeppers, line 18: So this introduces `$ANDROID_HOME` as a place for finding both the Android SDK & NDK, correct?

Are we worried this could break existing users? I could see users having `$ANDROID_HOME` set from Android Studio or something else, but Xamarin.Android is currently using the Android SDK installed by Visual Studio . A different SDK/NDK could get selected?

If we do this, should we put the env var checks last?

grendello: `$ANDROID_HOME` and `$ANDROID_SDK_ROOT` are checked, yes. Without these I wasn't able to make it detect the SDK instance I wanted it to use - it searched the `$PATH`. This addition will not affect users who set their preferred NDK location from the IDE, as that is still the first path checked. Furthermore, we now collect all the NDK paths and check their versions, then return the newest compatible one - that's the main goal of this PR. Environment variables should, generally, be considered before anything else, but it might break backward compatibility - it was an oversight in the original implementation of the code.

Note also that, currently, the order in which Windows and Unix look for the NDK is different:

  1. Preferred location set by the IDE (both Windows and Unix)
  2. Environment variables, with this PR (both Windows and Unix)
  3. Other locations
    1. `$PATH` directories (Unix)
    2. Known hard-coded locations (Windows)
  4. Other locations
    1. Known hard-coded locations (Unix, only mac really)
    2. `$PATH` directories (Windows)
  5. All discovered SDK locations (Windows)

This needs to be fixed in some future PR, but for now - not sure where the environment variable check would have to be placed? I feel it needs to be before `$PATH`, since that variable is usually set account or system wide (e.g. in system preferences for the user/machine on  Windows, and in `/etc` and user's shell init scripts on Unix) and it's not very comfortable to manipulate it ad-hoc, unlike the easy way to set the supported environment variables.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -230,6 +230,10 @@ protected override IEnumerable<string> GetAllAvailableAndroidNdks ()
 				if (CheckRegistryKeyForExecutable (root, regKey, MDREG_ANDROID_NDK, wow, ".", NdkStack))
 					yield return RegistryEx.GetValueString (root, regKey, MDREG_ANDROID_NDK, wow) ?? "";
 
+			foreach (string dir in GetSdkFromEnvironmentVariables ()) {
```
jonpryor, line 233: This is "dead code", as `GetAllAvailableAndroidNdks()` is only called from `AndroidSdkBase.GetValidNdkPath()`.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -94,6 +94,10 @@ protected override IEnumerable<string> GetAllAvailableAndroidSdks ()
 			if (!string.IsNullOrEmpty (preferedSdkPath))
 				yield return preferedSdkPath!;
 
+			foreach (string dir in GetSdkFromEnvironmentVariables ()) {
```
jonpryor, line 97: This is "dead code", as `GetAllAvailableAndroidNdks()` is only called from `AndroidSdkBase.GetValidNdkPath()`.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -139,7 +160,63 @@ protected virtual IEnumerable<string> GetAllAvailableAndroidNdks ()
...
+				foreach (string versionedNdkPath in Directory.EnumerateDirectories (ndkPath, "*", SearchOption.TopDirectoryOnly)) {
+					LoadNDKVersion (versionedNdkPath);
+				}
+			}
+
+			if (ndkInstances.Count == 0) {
+				return String.Empty;
+			}
+
+			return ndkInstances.First ().Value;
```
jonpryor, line 190: What does this *return*?  Won't it be the *lowest* found version?

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -139,7 +160,63 @@ protected virtual IEnumerable<string> GetAllAvailableAndroidNdks ()
 			foreach (var sdk in GetAllAvailableAndroidSdks ()) {
 				if (sdk == AndroidSdkPath)
 					continue;
-				yield return Path.Combine (sdk, "ndk-bundle");
+				yield return FindBestNDK (sdk);
+			}
+		}
+
+		string FindBestNDK (string androidSdkPath)
```
jonpryor, line 167: Nit-pick, and I'm not sure I should be saying this, *but*…

I'd prefer if this were "more pieces" with "clearer ordering":

```csharp
IEnumerable<string> GetSdkNdkPaths (string androidSdkPath)
{
    var ndk = Path.Combine (androidSdkPath, "ndk");
    if (Directory.Exists (ndk)) {
        foreach (string ndkPath in Directory.EnumerateDirectories (ndk, SearchOption.TopDirectoryOnly)) {
            yield return ndkDir;
        }
    }
    ndk = Path.Combine (androidSdkPath, "ndk-bundle");
    if (Directory.Exists (ndk))
        yield return ndk;
}

IEnumerable<string> OrderNdks (IEnumerable<string> ndkDirs)
{
    return from ndk in ndkDirs
        let ndkVersion = LoadNDKVersion (ndk)
        where ndkVersion > MaximumCompatibleNDKMajorVersion && ndkVersion < MinimumCompatibleNDKMajorVersion
        orderby ndkVersion descending
        select ndk;
}

string GetBestSdkNdkPath (string androidSdkPath)
{
    return OrderNdks (GetSdkNdkPaths (androidSdkPath)).FirstOrDefault ();
}
```

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
grendello: No, look at the instantiation of the `SortedDictionary` - it uses a comparer that sorts in descending order

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -139,7 +160,63 @@ protected virtual IEnumerable<string> GetAllAvailableAndroidNdks ()
...
+
+					if (!Version.TryParse (parts[1].Trim (), out Version? ndkVer) || ndkVer == null || ndkInstances.ContainsKey (ndkVer)) {
+						continue;
+					}
+
+					if (ndkVer.Major < MinimumCompatibleNDKMajorVersion || ndkVer.Major > MaximumCompatibleNDKMajorVersion) {
+						continue;
+					}
+
+					ndkInstances.Add (ndkVer, path);
```
jonpryor, line 217: What will this do with duplicates?  It throws `ArgumentException`.

It's possible that to have *both* an `ndk-bundle` dir *and* an `ndk/*` dir for the same version, which will cause this `.Add()` to throw.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
grendello: This is the code I've just added...

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
grendello: This is freshly added code and it's, eventually, called from `Initialize`

jonpryor:
This should also add a test to `tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkInfoTests.cs`.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
grendello: Let's not do this right now. This entire code needs to be revised and refactored - most of the stuff can be done in the base class, for both Unix and Windows (and should be). Let's postpone it for some near future

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
grendello: Look at the surrounding code - we check for duplicates.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
jonpryor: Never mind. sorry!

I thought `FindBestNDK()` called `GetSdkFromEnvironmentVariables()` as well, and was mistaken.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
jonpryor: Never mind. sorry!

I thought `FindBestNDK()` called `GetSdkFromEnvironmentVariables()` as well, and was mistaken.

grendello:
Please don't merge yet, working on adding one more test

grendello:
OK, ready

dellis1972 (approved):

jonathanpeppers (approved):

jonpryor:
```markdown
Context: https://github.com/actions/virtual-environments/issues/2420
Context: https://github.com/xamarin/xamarin-android/issues/5499
Context: https://github.com/xamarin/xamarin-android/issues/5526
Context: https://github.com/android/ndk/issues/1427
Context: https://developer.android.com/studio/command-line/variables#envar

Xamarin.Android is not (yet) compatible with the recently released
Android NDK r22 version.  Azure build images have recently rolled out
an update which includes NDK r22 and, thus, it breaks builds for
customers using any form of Xamarin.Android AOT build.

In an attempt to detect broken/incompatible NDK versions as well as
to select the "best one", this commit adds code to scan the known NDK
locations in search of the preferred version.

Given an `AndroidSdkInfo` creation of:

	var info = new AndroidSdkInfo (logger:logger,
	        androidSdkPath: sdkPath,
	        androidNdkPath: ndkPath,
	        javaSdkPath: javaPath);
	var usedNdkPath = info.AndroidNdkPath;

If `ndkPath` is not `null` and otherwise valid, then `usedNdkPath`
is `ndkPath`.

If `ndkPath` is `null` or is otherwise invalid (missing `ndk-stack`,
etc.), then we search for an `info.AndroidNdkPath` value as follows:

 1. If `androidSdkPath` is not `null` and valid, then we check for
    Android SDK-relative NDK locations, in:

      * `{androidSdkPath}/ndk/*`
      * `{androidSdkPath}/ndk-bundle`

    For each found SDK-relative NDK directory, we filter out NDKs for
    which we cannot determine the package version, as well as those
    which are "too old" (< `MinimumCompatibleNDKMajorVersion`) or
    "too new" (> `MaximumCompatibleNDKMajorVersion`), currently r22.

    We prefer the NDK location with the highest version number. 

 2. If there are no Android SDK-relative NDK locations, then we use
    the user-selected "preferred NDK location".
    See also `AndroidSdkInfo.SetPreferredAndroidNdkPath()`.

 3. If the preferred NDK location isn't set or is invalid, then we
    check directories specified in `$PATH`, and use the directory
    which contains `ndk-stack`.

 4. If no NDK location has been found, then we continue looking for
    NDK locations within the Android SDK locations specified by the
    `$ANDROID_HOME` and `$ANDROID_SDK_ROOT` environment variables.
    As with (1), these likewise look for e.g. `${ANDROID_HOME}/ndk/*`
    or `${ANDROID_SDK_ROOT}/ndk-bundle` directories and select the
    NDK with the highest supported version.

In (1) and 94), we now look for the Android SDK packages containing
the NDK.  There are two kinds of such packages:

  * `ndk-bundle` is the older package which allows for installation of
    only one NDK inside the SDK directory
  * `ndk/*` is a newer package which allows for installation of several
    NDK versions in parallel.  Each subdirectory of `ndk` is an `X.Y.Z`
    version number of the NDK.

In each of these directories we look for the `source.properties` file
from which we then extract the NDK version and then we sort thus
discovered NDK instances using their version as the key, in the
descending order.  The latest compatible (currently: less than 22 and
more than 15) version is selected and its path returned to the caller.
```

PR #108, opened 2021-02-24
[Xamarin.Android.Tools.AndroidSdk] Add API-S as a known version.

jpobst:
Add support for API-S to fix CI error:
```
 /Users/runner/Library/Android/dotnet/packs/Microsoft.Android.Sdk.osx-x64/11.0.200-ci.pr.gh5653.119/tools
/Xamarin.Android.Common.targets(517,2): error XA5207: Could not find android.jar for API level 31. This means
 the Android SDK platform for API level 31 is not installed. Either install it in the Android SDK Manager (Tools >
 Open Android SDK Manager...), or change the Xamarin.Android project to target an API version that is
 installed. (/Users/runner/Library/Android/sdk/platforms/android-31/android.jar missing.) [/Users/runner
/work/1/s/bin/TestRelease/temp/CheckLintErrorsAndWarnings/UnnamedProject.csproj]
```
when running One .NET tests.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs
@@ -182,6 +182,9 @@ namespace Xamarin.Android.Tools
 			new AndroidVersion (30, "11.0") {
 				AlternateIds = new[]{ "R" },
 			},
+			new AndroidVersion (31, "12.0") {
+				AlternateIds = new[]{ "S" },
+			},
 		};
 	}
 }


jonathanpeppers:
I think API 31 would would be able to work without this change. We have an `AndroidApiInfo.xml` file that describes all the API levels:

https://github.com/xamarin/xamarin-android/blob/2f83bddb99f6a0948631cd0f09bf037ba8a4e359/build-tools/create-packs/Microsoft.Android.Ref.proj#L44

Do we need an additional `AndroidApiInfo.xml` file in the `Microsoft.Android.Ref` package? Such as:

```
<AndroidApiInfo>
  <Id>31</Id>
  <Level>31</Level>
  <Name>S</Name>
  <Version>v12.0</Version>
  <Stable>False</Stable>
</AndroidApiInfo>
```

jonpryor:
This change shouldn't be necessary, and if it *is* necessary, we need to fix things to make it *not* necessary. :-)

PR #113, opened 2021-03-19
[Xamarin.Android.Tools.AndroidSdk] Probe for Microsoft OpenJDK dirs

jonpryor:


src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
```diff
@@ -313,6 +314,32 @@ static IEnumerable<string> GetConfiguredJdkPaths (Action<TraceLevel, string> log
...
+			return GetMacOSMicrosoftOpenJdkPaths ()
+				.Select (p => TryGetJdkInfo (p, logger, "/Library/Java/JavaVirtualMachines/microsoft-*.jdk"))
+				.Where (jdk => jdk != null)
+				.Select (jdk => jdk!)
+				.OrderByDescending (jdk => jdk, JdkInfoVersionComparer.Default);
+		}
+
+		static IEnumerable<string> GetMacOSMicrosoftOpenJdkPaths ()
+		{
+			var jdk     = AppDomain.CurrentDomain.GetData ($"GetMacOSMicrosoftOpenJdkPath override! {typeof (JdkInfo).AssemblyQualifiedName}")
```
mcumming, line 328: nit: spacing

src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
```diff
@@ -287,6 +287,7 @@ public static IEnumerable<JdkInfo> GetKnownSystemJdkInfos (Action<TraceLevel, st
 			return GetEnvironmentVariableJdks ("JI_JAVA_HOME", logger)
 				.Concat (GetWindowsJdks (logger))
 				.Concat (GetConfiguredJdks (logger))
+				.Concat (GetMacOSMicrosoftOpenJdks (logger))
 				.Concat (GetMacOSMicrosoftJdks (logger))
```
mcumming, line 291: We won't need this anymore either, correct? FWIK, this is for finding the Xamarin build of the JDK, which will no longer be shipped.

mcumming (changes requested):

src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
jonpryor: While true, I'm uncertain about the "compatibility", e.g. if someone installs a "newer" Xamarin.Android .vsix atop their "older" Visual Studio install.

Though even there, the odds are *really high* that `GetWindowsJdks()` > `GetPreferredJdkPaths()` (which hits the registry on Windows) or > `GetConfiguredJdks()` (which on macOS uses `monodroid-config.xml`) will return a value before `GetMacOSMicrosoftJdks()` will be encountered…

Thus sure, we can remove this.

grendello (approved):

src/Xamarin.Android.Tools.AndroidSdk/OS.cs
```diff
@@ -139,9 +140,89 @@ internal static class RegistryEx
 		static extern int RegCreateKeyEx (UIntPtr hKey, string subKey, uint reserved, string? @class, uint options,
 			uint samDesired, IntPtr lpSecurityAttributes, out UIntPtr phkResult, out Disposition lpdwDisposition);
 
+		// https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regenumkeyexw
+		[DllImport (ADVAPI, CharSet = CharSet.Unicode, SetLastError = true)]
+		static extern int RegEnumKeyExW (
```
jonathanpeppers, line 145: What is the need for p/invoke here? Is it an issue with 32-bit processes reading 64-bit keys?

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -193,6 +191,51 @@ private static IEnumerable<string> GetKnownOpenJdkPaths ()
...
+						continue;
+					paths.Add ((version, path));
+				}
+			}
+
+			return paths.OrderByDescending (e => e.version)
+				.Select (e => e.path);
+		}
+
+		static Version? ExtractVersion (string path, string prefix)
```
jonathanpeppers, line 216: Since this method does a bit of string manipulation, would it be worth writing a few unit tests that call it? To check you get the right result.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -163,28 +165,24 @@ private static IEnumerable<string> GetOpenJdkPaths ()
...
-				if (Directory.Exists (rootPath))  {
-					foreach (var directoryName in Directory.EnumerateDirectories (rootPath, $"{JdkFolderNamePattern}*").ToList ()) {
-						var versionString = directoryName.Replace ($"{rootPath}\\{JdkFolderNamePattern}", string.Empty);
-						if (Version.TryParse (versionString, out Version? ver)) {
-							paths.Add (new Tuple<string, Version>(directoryName, ver));
-						}
-					}
+				if (!Directory.Exists (rootPath))
+					continue;
+				foreach (var directoryName in Directory.EnumerateDirectories (rootPath, $"{JdkFolderNamePrefix}*").ToList ()) {
```
jonathanpeppers, line 181: Is the `.ToList()` needed? Seems like it will force the `EnumerateDirectories()` method to complete immediately.

src/Xamarin.Android.Tools.AndroidSdk/OS.cs
jonpryor: I'm not sure of the original need; this code *never* used `Microsoft.Win32.Registry` & co, and I'm not sure why.  (Apparently a question for @mhutch?)

Thus, given that we're already a tangled mess of P/Invokes, I figured it "best" to continue with the current paradigm.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
jonpryor: Good call; found at least one infinite loop doing that!

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
jonpryor: Nope; removed!

jonathanpeppers (approved):

mhutch:
@jonpryor wasn't me... https://github.com/xamarin/androidtools/commit/94e46bcc818b287a659ff8cd0d4c0ab83c80255b

PR #114, opened 2021-04-16
[Xamarin.Android.Tools.AndroidSdk] Probe for Zulu JDK

jonpryor:
Context: https://www.azul.com/downloads/zulu-community/

Commit 237642ce removed support for the ancient
`microsoft_dist_openjdk_*` versions, replacing it with the newer
[Microsoft OpenJDK][0].

One problem with this change is one of *compatibility*:
`microsoft_dist_openjdk_` was a build of JDK 1.8, while
Microsoft OpenJDK only provides builds for JDK 11 and later.
This is a problem, as the [Visual Studio Android Designer][1]
currently requires JDK 1.8, and is not yet compatible with JDK 11.

Add support to probe for for [Azul Zulu OpenJDK][2] installation
directories, as Zulu provides updated JDK 1.8 installers.

Additionally, refactor and centralize the JDK Location logic.
Previously, `JdkInfo.GetKnownSystemJdkInfos()` would return all of
the Windows installation locations before the macOS/Linux/etc.
locations, which made it possible for Windows and non-Windows
platforms to return paths in an inconsistent order.  Introduce a new
`JdkLocations` helper, and add new subclasses for each specific JDK
location, so that `JdkInfo.GetKnownSystemJdkInfos()` can be *unified*
across macOS/Linux & Windows, increasing consistency.

Add a new `tools/ls-jdks` tool, to make it easier to print all probed
JDK locations.

[0]: https://www.microsoft.com/openjdk
[1]: https://docs.microsoft.com/en-us/xamarin/android/user-interface/android-designer/designer-walkthrough?tabs=windows
[2]: https://www.azul.com/downloads/zulu-community/?package=jdk

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Jdks/OracleJdkLocations.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Jdks/OracleJdkLocations.cs
@@ -1,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Xamarin.Android.Tools {
+
+	class OracleJdkLocations : JdkLocations {
+
+		internal static IEnumerable<JdkInfo> GetOracleJdks (Action<TraceLevel, string> logger)
+		{
+			if (!OS.IsWindows) {
+				yield break;
+			}
+			foreach (var path in GetOracleJdkPaths ()) {
+				var jdk = JdkInfo.TryGetJdkInfo (path, logger, "Oracle Registry");
+				if (jdk == null) {
+					continue;
+				}
+				yield return jdk;
+			}
+		}
+
+		static IEnumerable<string> GetOracleJdkPaths ()
+		{
+			string subkey = @"SOFTWARE\JavaSoft\Java Development Kit";
+
+			foreach (var wow64 in new[] { RegistryEx.Wow64.Key32, RegistryEx.Wow64.Key64 }) {
+				string key_name = string.Format (@"{0}\{1}\{2}", "HKLM", subkey, "CurrentVersion");
+				var currentVersion = RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey, "CurrentVersion", wow64);
+
+				if (!string.IsNullOrEmpty (currentVersion)) {
+
+					if (RegistryEx.CheckRegistryKeyForExecutable (RegistryEx.LocalMachine, subkey + "\\" + "1.8", "JavaHome", wow64, "bin", "java.exe"))
+						yield return RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey + "\\" + "1.8", "JavaHome", wow64) ?? "";
+				}
+			}
+		}
+	}
+}
--- a/tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkWindowsTests.cs
+++ b/tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkWindowsTests.cs
@@ -13,56 +13,5 @@ namespace Xamarin.Android.Tools.Tests
 	[TestFixture]
 	public class AndroidSdkWindowsTests
 	{
-		[Test]
-		public void ExtractVersion ()
-		{
-			var sep = Path.DirectorySeparatorChar;
-
-			var tests = new[]{
-				new {
-					Path        = $"foo{sep}",
-					Prefix      = "",
-					Expected    = (Version) null,
-				},
-				new {
-					Path        = $"foo{sep}bar-1-extra",
-					Prefix      = "bar-",
-					Expected    = (Version) null,
-				},
-				new {
-					Path        = $"foo{sep}abcdef",
-					Prefix      = "a",
-					Expected    = (Version) null,
-				},
-				new {
-					Path        = $"foo{sep}a{sep}b.c.d",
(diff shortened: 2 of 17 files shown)


jonpryor:
TODO: write a better commit message.  (It's a draft!)

"Primary" motivation is to add support for the Azul Zulu JDK.

"Secondary" motivation -- which is *most* of the diff, by far -- is to bring a degree of sanity here.  Previously, `JdkInfo.GetKnownSystemJdkInfos()` was spread across Windows support, localized within `AndroidSdkWindows`, and non-Windows (macOS, Linux) support, within `JdkInfo` itself.

For me, at least, this makes understanding the lookups "weird".

*Unify* these systems, creating 6 "JDK Location" classes which know where to look for *one and only one* "JDK", for both macOS & Windows, and use these unified "finders" within `JdkInfo.GetKnownSystemJdkInfos()`.

I find this a tad "cleaner".

Keep in mind that we'll be wanting to add Yet Another JDK location next week, for Adopt OpenJDK, so I find making things "more understandable" to be a benefit.

grendello (commented):
LGTM, but I wonder if we should have a class that also looks in `~/android-toolchain/jdk` without having to put that location in `JAVA_HOME`? I know it would work only on systems where one builds Xamarin.Android, but it should be harmless on other machines while making it easier to use XA JDK as the main one on developer's machine.

src/Xamarin.Android.Tools.AndroidSdk/Jdks/OracleJdkLocations.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Xamarin.Android.Tools {
+
+	class OracleJdkLocations : JdkLocations {
+
+		internal static IEnumerable<JdkInfo> GetOracleJdks (Action<TraceLevel, string> logger)
+		{
+            if (!OS.IsWindows) {
+                yield break;
+            }
```
jonathanpeppers, line 16: I think there are some mixed tabs/spaces in here.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkWindowsTests.cs
```diff
@@ -13,56 +13,5 @@ namespace Xamarin.Android.Tools.Tests
 	[TestFixture]
 	public class AndroidSdkWindowsTests
 	{
-		[Test]
-		public void ExtractVersion ()
```
jonathanpeppers, line 17: Did this test get replaced by a new test?

src/Xamarin.Android.Tools.AndroidSdk/Jdks/OracleJdkLocations.cs
```diff
@@ -0,0 +1,49 @@
...
+
+					// No matter what the CurrentVersion is, look for 1.6 or 1.7 or 1.8
+					if (RegistryEx.CheckRegistryKeyForExecutable (RegistryEx.LocalMachine, subkey + "\\" + "1.8", "JavaHome", wow64, "bin", "java.exe"))
+						yield return RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey + "\\" + "1.8", "JavaHome", wow64) ?? "";
+
+					if (RegistryEx.CheckRegistryKeyForExecutable (RegistryEx.LocalMachine, subkey + "\\" + "1.7", "JavaHome", wow64, "bin", "java.exe"))
+						yield return RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey + "\\" + "1.7", "JavaHome", wow64) ?? "";
+
+					if (RegistryEx.CheckRegistryKeyForExecutable (RegistryEx.LocalMachine, subkey + "\\" + "1.6", "JavaHome", wow64, "bin", "java.exe"))
+						yield return RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey + "\\" + "1.6", "JavaHome", wow64) ?? "";
```
jonathanpeppers, line 44: Should we return 1.7 or 1.6 anymore? I don't know if those would even work.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkWindowsTests.cs
jonpryor: No; the method it was testing was removed, as the "let's extract a version code from a filesystem path" code was "replaced" with the (far older) "parse the version from the `release` file".

src/Xamarin.Android.Tools.AndroidSdk/Jdks/OracleJdkLocations.cs
```diff
@@ -0,0 +1,43 @@
...
+		{
+			string subkey = @"SOFTWARE\JavaSoft\Java Development Kit";
+
+			foreach (var wow64 in new[] { RegistryEx.Wow64.Key32, RegistryEx.Wow64.Key64 }) {
+				string key_name = string.Format (@"{0}\{1}\{2}", "HKLM", subkey, "CurrentVersion");
+				var currentVersion = RegistryEx.GetValueString (RegistryEx.LocalMachine, subkey, "CurrentVersion", wow64);
+
+				if (!string.IsNullOrEmpty (currentVersion)) {
+
+					// No matter what the CurrentVersion is, look for 1.6 or 1.7 or 1.8
```
jonathanpeppers, line 36: Comment is outdated

jonathanpeppers (approved):

dellis1972 (approved):

PR #129, opened 2021-07-28
[BaseTasks] improve Task settings in AsyncTaskExtensions

jonathanpeppers:
Fixes: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/1360410
Context: https://devblogs.microsoft.com/premier-developer/limiting-concurrency-for-faster-and-more-responsive-apps/

A work item came in from VS where AOT compilation is causing
responsiveness issues. The culprit came down to some of the
`TaskScheduler` and `TaskCreationOptions` settings in
`AsyncTaskExtensions`.

@davkean's recommendations are:

> This code needs to do two things:
> 1) It needs to pass `TaskCreationOptions.LongRunning`, so that we
>    don't burn queues meant for short lived tasks.
> 2) It needs to limit how many tasks it starts at once, to avoid CPU
>    contention and reduce the number of threads that we end up
>    burning.

Looking at `AsyncTaskExtensions` usage, it would happen during regular
builds as well, because aapt2-related tasks use these.

To improve these, I'm changing:

* New overloads to pass in `int maxConcurrencyLevel` and
  `TaskCreationOptions`.
* These default to `Environment.ProcessorCount * 2` and `LongRunning`.

When this change lands, we should potentially pass in
`$(Aapt2DaemonMaxInstanceCount)` where appropriate for
`maxConcurrencyLevel`.

I wrote a few tests to just check general sanity of
`AsyncTaskExtensions`. We didn't have any.

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Android.Build.Tasks
 			return pair.ConcurrentScheduler;
 		}
 
-		static int DefaultMaxConcurrencyLevel => Environment.ProcessorCount;
+		static int DefaultMaxConcurrencyLevel => Math.Max (1, Environment.ProcessorCount - 1);
 
 		static void LogErrorAndCancel (AsyncTask asyncTask, Exception exc)
 		{
--- a/src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
@@ -1,4 +1,4 @@
-﻿// https://github.com/xamarin/xamarin-android/blob/83854738b8e01747f9536f426fe17ad784cc2081/src/Xamarin.Android.Build.Tasks/Utilities/AsyncTaskExtensions.cs
+// https://github.com/xamarin/xamarin-android/blob/83854738b8e01747f9536f426fe17ad784cc2081/src/Xamarin.Android.Build.Tasks/Utilities/AsyncTaskExtensions.cs
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
```diff
@@ -79,28 +107,47 @@ public static ParallelLoopResult ParallelForEachWithLock<TSource> (this AsyncTas
...
+			TaskScheduler = GetTaskScheduler (maxConcurrencyLevel),
 		};
 
+		static TaskScheduler GetTaskScheduler (int maxConcurrencyLevel)
+		{
+			var pair = new ConcurrentExclusiveSchedulerPair (TaskScheduler.Default, maxConcurrencyLevel);
+			return pair.ConcurrentScheduler;
+		}
+
+		static int DefaultMaxConcurrencyLevel => Environment.ProcessorCount * 2;
```
davkean, line 121: I think you actually want to reduce this given the _LongRunning_. 

LongRunning == fire up a new worker thread just for the task, assuming AOT is CPU heavy, given a 4 core machine this will fire up 8 threads all competing for CPU. I think you'd be better off reducing this down to the same or just less than the ProcessorCount.

src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
```diff
@@ -79,28 +107,47 @@ public static ParallelLoopResult ParallelForEachWithLock<TSource> (this AsyncTas
...
+			TaskScheduler = GetTaskScheduler (maxConcurrencyLevel),
 		};
 
+		static TaskScheduler GetTaskScheduler (int maxConcurrencyLevel)
+		{
+			var pair = new ConcurrentExclusiveSchedulerPair (TaskScheduler.Default, maxConcurrencyLevel);
+			return pair.ConcurrentScheduler;
+		}
+
+		static int DefaultMaxConcurrencyLevel => Environment.ProcessorCount;
```
dellis1972, line 121: We might want to use a similar bit of logic as aapt2 here. See https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs#L82. 

We have a MAX number of daemons of 6 and a minimum of 1.

dellis1972:
> When this change lands, we should potentially pass in
$(Aapt2DaemonMaxInstanceCount) where appropriate for
maxConcurrencyLevel.

We should probably have separate properties for each, but unify them so the user can set both at once, or tweak each value. 
Something like

```xml
<AndroidMaxParallelBuildCount>6</AndroidMaxParallelBuildCount>
<Aapt2DaemonMaxInstanceCount Condition=" '$(Aapt2DaemonMaxInstanceCount)' == '' ">$(AndroidMaxParallelBuildCount )</Aapt2DaemonMaxInstanceCount>
<AotMaxBuildThreadCount Condition=" '$(AotMaxBuildThreadCount)' == '' ">$(AndroidMaxParallelBuildCount )</AotMaxBuildThreadCount >
```

(We'd need to work on the naming :)). 

Mind you , I not sure if anyone has ever used this property yet, apart from that one guy who one had 1 CPU and we ended up hanging the build because we calculated (at that time)  `ProcessorCount -1`  and started 0 daemons 🤦

src/Microsoft.Android.Build.BaseTasks/AsyncTaskExtensions.cs
jonathanpeppers: What if this was:
```suggestion
		static int DefaultMaxConcurrencyLevel => Math.Max (1, Environment.ProcessorCount - 1);
```

I think about AOT or other random tasks where we do `Parallel.ForEach()`, I'd think we'd want it to use more than 6 on my PC, for example. It has 16 cores.

`aapt2` is a little different because it's starting long running daemons. Where I think 6 would be sufficient on a many-core machine.

jonathanpeppers:
Since they can't control AOT parallelization now, I think we could probably just choose a default here and not need to create new properties.

dellis1972 (approved):

jonathanpeppers:
I saw this API that `@radical` was using here:

https://github.com/dotnet/runtime/pull/59391/files#diff-2f7f7e10f4474e62ffedeb8c4cca26556078e3b977b1bd73b1ecd7ec7c0fda04R384

```csharp
if (BuildEngine is IBuildEngine9 be9)
    allowedParallelism = be9.RequestCores(allowedParallelism);
```

Maybe we should actually use this?

dellis1972:
Interesting. We'd need to find out if that is for MSBuild Project Parallelism or if we can use it for our Background Thread stuff.

PR #130, opened 2021-08-11
[Xamarin.Android.Tools.AndroidSdk] More Microsoft Dist JDK Support

jonpryor:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1314263

Remember `AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()`
(commit a4aad186)?  It's supposed to "detect" the latest installed
"preferred" JDK installation location, and set configuration values
so that subsequent `new AndroidSdkInfo()` expressions will use that
JDK location, instead of some other JDK location.

The "preferred" JDK was a JDK that was "known good" for Xamarin.Android
use, i.e. one that Visual Studio installed.  In a4aad186, this was the
`microsoft_dist_openjdk_*` version.

When we (prematurely) removed support for the `microsoft_dist_openjdk_`
build in commits 237642ce and dca30d9d, we updated
`AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()` to instead
prefer the [Microsoft OpenJDK][0] build, but this migration needed to
be delayed until Visual Studio 17.0.

Which means that the "current world order" is *wrong*:
`AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()` tries to
detect a JDK that is not currently installed by anything.

Futhermore, Visual Studio for Mac will call
`AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()` if no JDK
is currently configured.

The result is that on "clean" systems, Visual Studio for Mac can
crash with a `NotSupportedException` because it's looking for a
Microsoft OpenJDK installation, but only a `microsoft_dist_openjdk_`
installation is available.  Thus, no preferred JDK is found, and the
IDE crashes.

Fix this by adding a new (`internal`) `JdkInfo.GetPreferredJdkInfos()`
method, which appropriately checks *both* Microsoft OpenJDK *and*
`microsoft_dist_openjdk_` installation locations, and update
`AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()` to now
use `JdkInfo.GetPreferredJdkInfos()`.

(We add the method to `JdkInfo` so that it is easier to keep
`GetPreferredJdkInfos()` and `GetKnownSystemJdkInfos()` consistent
with each other.)

[0]: https://www.microsoft.com/openjdk

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -185,7 +185,7 @@ namespace Xamarin.Android.Tools
 
 			logger          = logger ?? DefaultConsoleLogger;
 
-			var latestJdk   = MicrosoftOpenJdkLocations.GetMicrosoftOpenJdks (logger).FirstOrDefault ();
+			var latestJdk   = JdkInfo.GetPreferredJdkInfos (logger).FirstOrDefault ();
 			if (latestJdk == null)
 				throw new NotSupportedException ("No Microsoft OpenJDK could be found.  Please re-run the Visual Studio installer or manually specify the JDK path in settings.");
 
--- a/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
@@ -280,6 +280,7 @@ namespace Xamarin.Android.Tools
 			return props;
 		}
 
+		// Keep ordering in sync w/ GetPreferredJdkInfos
 		public static IEnumerable<JdkInfo> GetKnownSystemJdkInfos (Action<TraceLevel, string>? logger = null)
 		{
 			logger  = logger ?? AndroidSdkInfo.DefaultConsoleLogger;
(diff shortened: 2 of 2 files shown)


jonathanpeppers (approved):
The fix looks good to me.

Down the road, is there something we could do to test these changes better? Would mocking the file system, environment variable, and/or registry calls be doable?

jonpryor:
We do have some code to create a "faux" JDK: https://github.com/xamarin/xamarin-android-tools/blob/f9c1b0d754050cef0a8339ff87a37882fa13a9f6/tests/Xamarin.Android.Tools.AndroidSdk-Tests/JdkInfoTests.cs#L60-L100

We also have support for "overriding" the default global locations: https://github.com/xamarin/xamarin-android-tools/blob/f9c1b0d754050cef0a8339ff87a37882fa13a9f6/src/Xamarin.Android.Tools.AndroidSdk/Jdks/MicrosoftDistJdkLocations.cs#L21

(Not "complete" support for overriding, but it's "something".)

My "problem" with a unit test here is that it wouldn't help with the original bug: the unit test would just assert that the method does what it does, but *wouldn't* help ensure that `AndroidSdkInfo.DetectAndSetPreferredJavaSdkPathToLatest()` is doing what it's *supposed* to do.

Rephrased: even if there were tests, we'd likely have overlooked those tests as part of the whole "remove support for `microsoft_dist_openjdk_`, oh wait bring it back, oh wait…" rigamarole that we went through this spring.

PR #135, opened 2021-09-23
Allow targets override

BretJohnson:
Allow a .targets.override file in the parent directory
to customize targets settings for Directory.build.targets.

This is the same change made for androidtools here
https://github.com/xamarin/androidtools/pull/307,
applied to xamarin-android-tools.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -13,4 +13,9 @@
       <InformationalVersion>$(Version); git-rev-head:$(GitCommit); git-branch:$(GitBranch)</InformationalVersion>
     </PropertyGroup>
   </Target>
+
+  <Import
+      Project="$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory))).override.targets"
+      Condition=" Exists('$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory))).override.targets') "
+  />
 </Project>


SvetBonev (approved):

jonathanpeppers (approved):

jonpryor:
…but what properties do you need to override?  There's no Git submodules, which was the primary use case in the other submodules.

jonpryor:
> …but what properties do you need to override?

Not quite a relevant question; this isn't an extension point for `Directory.Build.props`, it's an extension point for *`Directory.Build.targets`*.

See also: https://github.com/xamarin/XamarinVS/pull/12550

Directory.Build.targets
```diff
@@ -13,4 +13,9 @@
       <InformationalVersion>$(Version); git-rev-head:$(GitCommit); git-branch:$(GitBranch)</InformationalVersion>
     </PropertyGroup>
   </Target>
+
+  <Import
```
jonpryor, line 17: Question: should this Import be *last*, as it is here? Or *first*?

BretJohnson: It needs to be last - I had it first before and it didn't work. The primary use case here is to redefine the SetVersion target to be something else, so it needs to come after the SetVersion target to override it.

BretJohnson:
@gitbot backport d17-0

PR #144, opened 2021-11-10
Update MaximumCompatibleNDKMajorVersion to be 23

dellis1972:
Fixes https://work.azdo.io/1433453

Users are reporting the following error when trying to use NDK 22

```
MSB4044: The "MakeBundleNativeCodeExternal" task was not given a value for the required parameter "AndroidNdkDirectory".
```

Earlier in the logs we see this message

```
Skipping NDK in 'C:\ProgramData\Android\android-sdk
dk\22.1.7171670': version 22.1.7171670 is out of the accepted range (major version must be between 16 and 21
```

We need to update the `MaximumCompatibleNDKMajorVersion` to match our new
upper limit.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Android.Tools
 	{
 		// When this changes, update the test: Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Ndk_MultipleNdkVersionsInSdk
 		const int MinimumCompatibleNDKMajorVersion = 16;
-		const int MaximumCompatibleNDKMajorVersion = 22;
+		const int MaximumCompatibleNDKMajorVersion = 23;
 
 		static readonly char[] SourcePropertiesKeyValueSplit = new char[] { '=' };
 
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
```diff
@@ -11,7 +11,7 @@ abstract class AndroidSdkBase
 	{
 		// When this changes, update the test: Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Ndk_MultipleNdkVersionsInSdk
 		const int MinimumCompatibleNDKMajorVersion = 16;
-		const int MaximumCompatibleNDKMajorVersion = 21;
+		const int MaximumCompatibleNDKMajorVersion = 22;
```
grendello, line 14: Should we bump it to 23? That's what XA/main builds with now.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkInfoTests.cs
```diff
@@ -71,7 +71,7 @@ public void Ndk_MultipleNdkVersionsInSdk ()
 		{
 			// Must match like-named constants in AndroidSdkBase
 			const int MinimumCompatibleNDKMajorVersion = 16;
-			const int MaximumCompatibleNDKMajorVersion = 21;
+			const int MaximumCompatibleNDKMajorVersion = 22;
```
grendello, line 74: Likewise here

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkInfoTests.cs
```diff
@@ -91,6 +91,7 @@ public void Ndk_MultipleNdkVersionsInSdk ()
 				"21.2.6472646",
 				"21.3.6528147",
 				"22.0.7026061",
+				"22.1.7171670",
```
grendello, line 94: We now use `23.1.7779620` (NDK r23b)

PR #149, opened 2022-01-24
Bump LibZipSharp to 2.0.2

dellis1972:
Changes

- Use a different license tag for nuget generation 
- Fix corrupt data when Deleting entries from zip files.

https://github.com/xamarin/LibZipSharp/compare/2.0.0...2.0.2


jonathanpeppers (approved):
Can you add a link to the xamarin/libZipSharp diff in the commit message?

Otherwise looks good.

PR #157, opened 2022-02-17
Don't return null Jdk locations

KirillOsenkov:
VSMac has a first-chance exception because we're returning null here for some attributes

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Jdks/JdkLocations.MacOS.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Jdks/JdkLocations.MacOS.cs
@@ -21,8 +21,10 @@ namespace Xamarin.Android.Tools {
 		{
 			var config = AndroidSdkUnix.GetUnixConfigFile (logger);
 			foreach (var java_sdk in config.Root.Elements ("java-sdk")) {
-				var path    = (string) java_sdk.Attribute ("path");
-				yield return path;
+				var path = (string) java_sdk.Attribute ("path");
+				if (path != null) {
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
My xml looks like this:

image

src/Xamarin.Android.Tools.AndroidSdk/Jdks/JdkLocations.MacOS.cs
```diff
@@ -21,8 +21,10 @@ static IEnumerable<string> GetUnixConfiguredJdkPaths (Action<TraceLevel, string>
 		{
 			var config = AndroidSdkUnix.GetUnixConfigFile (logger);
 			foreach (var java_sdk in config.Root.Elements ("java-sdk")) {
-				var path    = (string) java_sdk.Attribute ("path");
-				yield return path;
+				var path = (string) java_sdk.Attribute ("path");
```
jonpryor, line 24: Nullable reference types are enabled, and the original code -- as well as this updated code -- is "disabling" NRT by asserting that `string` and not `string?` is returned.

KirillOsenkov: I opted for the minimal surgical fix to solve the problem since I'm in an unfamiliar codebase. I can leave more comprehensive fix options to the maintainers. Thanks!

jonpryor:
@KirillOsenkov: i consider this PR to be superseded by PR #158, but I would love to have a version of the first-chance exception that VSMac is seeing for the #158 commit message.

PR #171, opened 2022-06-14
[Xamarin.Android.Tools.AndroidSdk] Permit NDK r24

jonpryor:
Update `AndroidSdkBase.MaximumCompatibleNDKMajorVersion` to 24, so
that NDK r24 is considered as a valid version.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs
@@ -11,7 +11,7 @@ namespace Xamarin.Android.Tools
 	{
 		// When this changes, update the test: Xamarin.Android.Tools.Tests.AndroidSdkInfoTests.Ndk_MultipleNdkVersionsInSdk
 		const int MinimumCompatibleNDKMajorVersion = 16;
-		const int MaximumCompatibleNDKMajorVersion = 23;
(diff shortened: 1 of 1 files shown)


grendello (approved):
It's fine for now, but I think this value should be settable by the users of the library (XA, IDEs etc)

pjcollins (approved):

pjcollins:
Added https://github.com/xamarin/xamarin-android-tools/issues/172 to track @grendello's comment.

jonpryor:
@grendello: do we need a "max supported NDK" at all, *here* (in xamarin-android-tools)?  It's not a public member, and thus nothing else can easily use it for anything, *and* it's not even used to verify constructor parameters, which will be specified "elsewhere" ( https://github.com/xamarin/xamarin-android/blob/944f88a8b6f1c30eaf43f56cfbebf0f8cd7a547f/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs#L122 ).

Why does `AndroidSdkInfo` -- and not even all codepaths!  just "we' re looking around randomly for an SDK & NDK" codepaths! -- need to check that the detected NDK is within a supported range?

grendello:
@jonpryor it was added to prevent us from using NDKs we don't support (e.g. we didn't have support for the clang-only NDKs for a while) on the end user's machine. The idea was to yell loud if the user updated to latest NDK that we have no support for yet.

Issue #140, opened 2021-10-19
Missing mapping for all known Android Versions and corresponding code names and API levels 


pjcollins:
Xamarin.Android is currently responsible for generating and installing `AndroidApiInfo.xml` files alongside `Mono.Android.dll` to help provide a full mapping of Android version information. For reference, these files look like this:

```xml
<AndroidApiInfo>
  <Id>30</Id>
  <Level>30</Level>
  <Name>R</Name>
  <Version>v11.0</Version>
  <Stable>True</Stable>
</AndroidApiInfo>
```

While these files are great at providing context around the Android versions that a current installation of Xamarin.Android supports, they do not provide a full story with respect to all known Android versions. As a fall back, we have this hardcoded mapping:
https://github.com/xamarin/xamarin-android-tools/blob/f2af06f2a2c1966011b6b04a113560de37041cbc/src/Xamarin.Android.Tools.AndroidSdk/AndroidVersions.cs#L142

jonpryor:
The reason it's in xamarin-android-tools is because androidtools *also* needs the data, plus various other dependencies.  We thus *want* xamarin-android to be the "source of truth", but *don't* want it to provide the *API* for the source of truth.  (Which is kinda how it works now: xamarin-android creates the `AndroidApiInfo.xml` files, while xamarin-android-tools *reads* those files to implement `AndroidVersions`.)

But wait, it gets different/worse!  *Part* of the intent behind the use of `AndroidApiInfo.xml` files was so that the IDE could be "dynamic", and automagically "pick up" new API levels without requiring code changes.

As originally conceived, there would be one `AndroidApiInfo.xml` file per `$(TargetFrameworkVersion)` value.

The problem is that in our forthcoming .NET 6+ world order, there is only "one" `$(TargetFrameworkVersion)` value, not 13+.

We thus need a "re-conception" of how this *should* work, ideally in a manner which preserves the current system's "dynamism" (which we're gonna need in ~4-6 months once API-32 previews are available…).  A `KnownAndroidApiInfo.xml` file would be such an idea, but we also need to figure out how to *find* it in a manner which works everywhere it's needed.

jonpryor:
Might also be handy if the consolidated version info also could contain the highest supported NDK version.

Issue #61, opened 2019-01-30
"Revert" PR #57


jonpryor:
We need to better understand why `AndroidVersions.KnownVersions` needs to be public, because making it public is almost certainly the *wrong* thing to do, long-term.

In particular, the PR message is incorrect:

> KnownVersions should be the only place we modify when new versions are published.

This is wrong; see also commit aad97f165317a6308dd34aeca33998c894743804.  The intention is that *`xamarin-android`* should control which versions are "supported," via the presence of the `AndroidApiInfo.xml` files.  Using the `AndroidApiInfo.xml` files is *key*, so that we can plausibly introduce *new* API levels (API-Q, anyone?) *without* requiring *code* changes to `xamarin-android-tools` (and the requisite product bumps across 6+ repos).

In short, *what is the **actual** need* behind PR #57?

> For Android Manifest we also need to know what exists because Minimum Android Version and Target may be anything

This is in fact not correct either; there *are* limitations.  Case in point: https://github.com/xamarin/xamarin-android/commit/06e920c18b71a165eded0a5782f1a285ef20a4bc.

NDK r19 -- which xamarin-android/master is now using -- has a *minimum* API level of 16...*which varies across ABI targets*.  `armeabi-v7a` and `x86` have a minimum API level of 16, while `arm64-v8a` and `x86_64` have minimum API levels of 21.  Trying to use a lower value *may* work, but it's highly likely they *won't* work, because of e.g. native linker symbol references.

In any event, anything prior to API-16 won't be QA'd or supported.

Which is a long-winded way of saying that this problem space is more complicated, and we need to come up with a more complete (and updatable!) solution.

Just making `AndroidVersions.KnowVersions` public Is Not The Correct Approach.

PR #183, opened 2022-07-25
[Xamarin.Android.Tools.AndroidSdk] Add API-33 to KnownVersions

tondat:


jonpryor:
I would still like to know *why* this change is necessary.  See also:

  * https://github.com/xamarin/xamarin-android-tools/issues/61
  * https://github.com/xamarin/xamarin-android-tools/issues/140

We should not *need* this PR.  If we do, something is broken, and we should fix it.

PR #188, opened 2022-08-02
Fallback to well known config file place

rodrmoya:
When running on NET6 (VSMac installer), SpecialFolder.ApplicationData
returns nothing, resulting in a relative path being used for the config
file. In those cases, check that's the case and fallback to well known
~/.config folder, which is, from what it seems, the default.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
@@ -221,10 +221,10 @@ namespace Xamarin.Android.Tools
 		//
 		private static string UnixConfigPath {
 			get {
-				var p   = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
+				var p = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
 				if (string.IsNullOrEmpty (p)) {
-					p   = Environment.GetFolderPath (Environment.SpecialFolder.ApplicationData);
-					p   = Path.Combine (p, "xbuild");
+					p = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.UserProfile), ".config");
+					p = Path.Combine (p, "xbuild");
 				}
 				return Path.Combine (p, "monodroid-config.xml");
 			}
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
@@ -221,10 +221,10 @@ namespace Xamarin.Android.Tools
 		//
 		private static string UnixConfigPath {
 			get {
-				var p = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
+				var p   = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -221,10 +221,16 @@ void AddToList (string path)
 		//
 		private static string UnixConfigPath {
 			get {
-				var p   = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
+				var p = AppDomain.CurrentDomain.GetData (GetUnixConfigDirOverrideName)?.ToString ();
 				if (string.IsNullOrEmpty (p)) {
 					p   = Environment.GetFolderPath (Environment.SpecialFolder.ApplicationData);
```
jonathanpeppers, line 226: What value does this method return under Mono? Should we just change it to:
```suggestion
					p   = Environment.GetFolderPath (Environment.SpecialFolder.UserProfile);
```

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -221,10 +221,16 @@ void AddToList (string path)
...
-					p   = Path.Combine (p, "xbuild");
+					if (string.IsNullOrEmpty(p))
+					{
+						// SpecialFolder.ApplicationData returns nothing when running on NET6, so in that
+						// case, fallback to the well known place
+						p = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".config");
+					}
```
jonathanpeppers, line 232: If we still need this `if`, we normally use this code style:

```suggestion
					if (string.IsNullOrEmpty (p))
					{
						// SpecialFolder.ApplicationData returns nothing when running on NET6, so in that
						// case, fallback to the well known place
						p = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.UserProfile), ".config");
					}
```

It's in the `.editorconfig`, maybe this setting isn't working?

https://github.com/xamarin/xamarin-android-tools/blob/0d55472777ebf1a38b72a87c5bd3f72ecafa4832/.editorconfig#L183

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
rodrmoya: On Mono it's doing the correct thing, this same code works perfectly on the mono-based Installer, and see:
```
$ csi
Microsoft (R) Visual C# Interactive Compiler version 3.9.0-6.21124.20 ()
Copyright (C) Microsoft Corporation. All rights reserved.

Type "#help" for more information.
> Environment.GetFolderPath (Environment.SpecialFolder.ApplicationData);
> var p = Environment.GetFolderPath (Environment.SpecialFolder.ApplicationData);
> Console.WriteLine(p);
/Users/rodrigo/.config
```
that's why I added this code as a fallback, to not break non-net6 scenarios.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
rodrmoya: Fixed in https://github.com/xamarin/xamarin-android-tools/pull/188/commits/58a1c3bb038751deca2ccf6ea0aa08cec6d5c4ac

jonathanpeppers (approved):

PR #194, opened 2022-09-13
Fix SpecialFolder.Personal usages

eerhardt:
All of the usages of SpecialFolder.Personal should be using UserProfile instead. SpecialFolder Personal is the same value as MyDocuments, both of which will be changed on Unix from `~` to `~/Documents` in .NET 8. See https://github.com/dotnet/runtime/pull/68610.


eerhardt:
cc @jonathanpeppers @jonpryor

jonathanpeppers (approved):

PR #195, opened 2022-10-11
Move from `netcoreapp3.1` to `net6.0`

jonathanpeppers:
I believe this is a good idea anyway, but this should assist with:

https://github.com/xamarin/xamarin-android/pull/7451

Which is hitting the error:

    error NU1102: Unable to find package Microsoft.NETCore.App.Host.osx-x64 with version (= 3.1.31)

.NET Core 3.1.31 is not released yet and is likely on a private feed.

Also pass `6.0.x` to the [`UseDotNet`][0] yaml task.

After these changes, the build failed with:

    error MSB3971: The reference assemblies for ".NETFramework,Version=v6.0" were not found. You might be using an older .NET SDK to target .NET 5.0 or higher. Update Visual Studio and/or your .NET SDK.

Mono MSBuild probably can't build .NET 6 projects?

So I fixed various things in the build:

* Use `dotnet build` instead of MSBuild, remove NuGet commands
* No tests were running:
```
[warning] Project file(s) matching the specified pattern were not found.
```
* Fixed the `$(XATBuildingForNetCoreApp)` property
* Upload artifacts based on the pool name: `$(vmImage)`
* Use `$(PackageOutputPath)` instead of `$(OutputPath)`
* `PublishBuildArtifacts` is deprecated use `PublishPipelineArtifact`
* Save `.binlog` files as artifacts

[0]: https://learn.microsoft.com/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer

### DIFF
--- a/azure-pipelines.yaml
+++ b/azure-pipelines.yaml
@@ -12,7 +12,7 @@ pr:
 
 # Global variables
 variables:
-  DotNetCoreVersion: 3.1.100
+  DotNetCoreVersion: 6.0.x
 
 jobs:
 - job: build
@@ -42,22 +42,8 @@ jobs:
     inputs:
       version: $(DotNetCoreVersion)
 
-  - task: NuGetToolInstaller@0
-    displayName: 'Install NuGet'
-    inputs:
-      versionSpec: 5.x
-
-  - task: NuGetCommand@2
-    displayName: 'NuGet Restore'
-    inputs:
-      restoreSolution: Xamarin.Android.Tools.sln
-      feedsToUse: config
-      nugetConfigPath: NuGet.config
-
-  - task: MSBuild@1
+  - script: dotnet build Xamarin.Android.Tools.sln -bl:$(Build.ArtifactStagingDirectory)/build.binlog
     displayName: 'Build solution Xamarin.Android.Tools.sln'
-    inputs:
-      solution: Xamarin.Android.Tools.sln
 
   - task: DotNetCoreCLI@2
     displayName: 'Run Tests'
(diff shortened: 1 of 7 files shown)


azure-pipelines.yaml
```diff
@@ -42,28 +42,14 @@ jobs:
...
-
-  - task: NuGetCommand@2
-    displayName: 'NuGet Restore'
-    inputs:
-      restoreSolution: Xamarin.Android.Tools.sln
-      feedsToUse: config
-      nugetConfigPath: NuGet.config
-
-  - task: MSBuild@1
+  - script: dotnet build Xamarin.Android.Tools.sln -bl:$(Build.ArtifactStagingDirectory)/build.binlog
```
pjcollins, line 45: It may be preferable to use DotNetCoreCLI@2 here, though it shouldn't make much of a difference

```yaml
  - task: DotNetCoreCLI@2
    displayName: 'Build solution Xamarin.Android.Tools.sln'
    inputs:
      projects: Xamarin.Android.Tools.sln
      arguments: -bl:$(Build.ArtifactStagingDirectory)/build.binlog
```

azure-pipelines.yaml
```diff
@@ -73,15 +59,13 @@ jobs:
       Write-Host "##vso[task.setvariable variable=xat.nuget.version]$version"
     condition: and(succeeded(), eq(variables['agent.os'], 'Windows_NT'))
 
-  - task: MSBuild@1
+  - script: dotnet pack src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj -p:Version=$(xat.nuget.version) -p:PackageOutputPath=$(Build.ArtifactStagingDirectory) -bl:$(Build.ArtifactStagingDirectory)/pack.binlog
```
pjcollins, line 62: We could also use `DotNetCoreCLI@2` here instead.

azure-pipelines.yaml
jonathanpeppers: I thought I would also have to put `command: build` or `command: pack`.

Does it *do* anything extra that is helpful? Like insert warnings inline in PRs? Something else?

Otherwise a one-line `script:` block seems simpler?

pjcollins:
@tondat is there anything we need to consider on the IDE side of things for this?

jonathanpeppers:
> @tondat is there anything we need to consider on the IDE side of things for this?

I think the IDEs would all use the `netstandard2.0` assemblies, such as:
```xml
<TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```

azure-pipelines.yaml
pjcollins: The `command: build` param is used by default, so it doesn't need to be set when building. You would need to set `command: pack` for packing though.

I think the only advantage to using the task over a script is that it will probe multiple locations for `dotnet`, but since we're using the `UseDotNet` task there shouldn't be a difference either way.

pjcollins (approved):

tondat (approved):
Please try to verify the build of androidtools repo with the bump of this change

jonathanpeppers:
I will test the latest in androidtools & xamarin-android shortly.

jonathanpeppers:
FYI @tondat seemed to work fine: https://github.com/xamarin/androidtools/pull/362

PR #203, opened 2023-02-16
Bump 'AndroidSdkPlatformToolsVersion' to 34.0.0

dellis1972:


jonpryor:
We cannot bump this default change until android-sdk-installer likewise supports platform-tools 34.0.0.

@tondat: who's backlog can we put this on? :-)

tondat:
@jonpryor working on that. 

The idea is update the feed for 17.6 ?

tondat:
@jonpryor this is the proposed feed with the changes: https://github.com/xamarin/android-sdk-installer/pull/765
if all is ok we can update the feed.

tondat:
@jonpryor these PR has the change on willow: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/454760

and this PR has the proposed new manifest https://github.com/xamarin/android-sdk-installer/pull/765

waiting for your approval to proceed.

jonpryor:
@tondat: so here's the philosophical conundrum: *API-34* is still in preview, and won't be stable [until July or later](https://developer.android.com/about/versions/14/overview).  However, that's the *platform* package; here, we're talking about the *platform-tools* package, and *here* we have a bunch of unknowns:

 1. Why is platform-tools 34.0.0 already stable?  (More rhetorical than anything, but this is *really* weirding me out.)
 2. Do we work with platform-tools 34.0.0?  As far as I know, we have *not* attempted to to go through the xamarin-android unit tests with platform-tools 34.0.0, and "*normally*" I would expect this to happen as part of [API-34 binding](https://github.com/xamarin/xamarin-android/pull/7796), which is still in-progress.  @jpobst? :-D

Given (2), we should *not* update Willow to install *only* platform-tools 34.0.0 now, and possibly not until API-34 is the *default* (2023- *November*?).

xamarin/android-sdk-installer#765 should be "fine" (-ish; I have some comments to leave there), as it is only *additive*, not replacing.

Which still leaves *this* PR: we shouldn't merge it until we're able to make platform-tools 34.0.0 the default *everywhere*.  *Maybe* that's "sooner" -- once we've done a complete test pass using platform-tools 34.0.0 and have verified it works with API-31…API-33 apps -- or maybe that's November, but it's certainly not *now*.

jonpryor:
@jpobst [points out](https://discord.com/channels/732297728826277939/732297837953679412/1080177167931035688) that I'm overthinking this: new platform-tools versions regularly come out in February; see the [SDK Platform Tools release notes](https://developer.android.com/studio/releases/platform-tools):

> * 34.0.0 (February 2023)
> * 33.0.0 (February 2022)
> * 32.0.0 (January 2022)
> * 31.0.0 (February 2021)

I am thus freaking out more than is strictly necessary, and this bump is likely safe.

That said, we should still try a bump in xamarin-android first, just to verify that nothing breaks.  Nothing *should* break, but there is no harm to being careful here.

jonpryor:
On the "Does xamarin-android Build™ with platform-tools 34.0.0" front, the answer is a tentative "Yes":

  * xamarin-android/main is Green: https://github.com/xamarin/xamarin-android/pull/7863
  * xamarin-andorid/d17-5 is mostly green? https://github.com/xamarin/xamarin-android/pull/7854

If both of those PRs are fully green, then *Yes*, we work with platform-tools 34.0.0.

---

*That said*… if the intention is also to change [*the VS installer*](https://github.com/xamarin/xamarin-android-tools/pull/203#issuecomment-1448467024), then we will *also* need to apply the changes to `$(AndroidSdkPlatformToolsVersion)` to all the relevant d17-5 branches, including (but not limited to?):

  * xamarin-android-tools/d17-5
  * xamarin-android/d17-5
  * monodroid/d17-5
  * androidtools/d17-5
  * android-sdk-installer/d17-5 (???)

@tondat: is that a set of changes you are comfortable with?  *Or* do we need to consider "bumping" monodroid/d17-5 to use "newer" androidtools & android-sdk-installer branches?

(Background: xamarin-android/main is no longer completely building Classic Xamarin.Android; xamarin-android/d17-5 is our ongoing maintenance branch for Classic Xamarin.Android. We *could* plausibly create a set of d17-6 branches from d17-5 -- for xamarin-android-tools, xamarin-android, monodroid -- but this is something we need to discuss.)

jonpryor:
Platform-tools 34.0.1 is current latest version on https://dl-ssl.google.com/android/repository/repository2-3.xml

PR #206, opened 2023-05-09
[Xamarin.Android.Tools.AndroidSdk] Fix NRT warnings.

jpobst:


src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
```diff
@@ -21,9 +21,9 @@ public class JdkInfo {
 
 		public      string?                             Locator                     {get;}
 
-		public      string                              JarPath                     {get;}
-		public      string                              JavaPath                    {get;}
-		public      string                              JavacPath                   {get;}
+		public      string?                             JarPath                     {get;}
```
jonpryor, line 24: I don't like these changes, and am not sure why they're needed.  For example, if this is "*actually*" a JDK, then *surely* it has `jar` and `javac` and `java`, so these *must* exist. (No?)  Additionally, these don't *currently* elicit warnings, at least not for me.  Thus, why are these changes needed?

jpobst: We explicitly set these values to `null` when they cannot be found:
https://github.com/xamarin/xamarin-android-tools/blob/main/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs#L64-L66

This results in the following warnings without these changes:
```
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(64,26,64,100): warning CS8601: Possible null reference assignment.
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(65,26,65,101): warning CS8601: Possible null reference assignment.
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(66,26,66,102): warning CS8601: Possible null reference assignment.
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(52,10,52,17): warning CS8618: Non-nullable property 'JarPath' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(52,10,52,17): warning CS8618: Non-nullable property 'JavaPath' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
1>C:\code\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\JdkInfo.cs(52,10,52,17): warning CS8618: Non-nullable property 'JavacPath' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
```

jonpryor: I wasn't properly reading my warning messages. :-(
Was expecting something on line 24. :-/

That said, they *are* required!

https://github.com/xamarin/xamarin-android-tools/blob/main/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs#L70-L73

It's just that they're "required" in a form that the C# compiler doesn't know about.

So maybe instead we should do:

```csharp
JarPath = RequireExecutableInDirectory (binPath, "jar");

// …
string RequireExecutableInDirectory (string binPath, string filename)
{
    var file = ProcessUtils.FindExecutablesInDirectory (binPath, filename).FirstOrDefault ();
    if (file == null)
        throw new ArgumentException ($"Could not find required file `{filename}` in {HomePath}; …");
    return file;
}
```

*That* should appease the C# compiler, right?

jpobst: 👍 Makes sense, Fixed it roughly as suggested.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs
```diff
@@ -158,6 +166,10 @@ public override void SetPreferredAndroidNdkPath (string? path)
 			path = NullIfEmpty (path);
 
 			var doc = GetUnixConfigFile (Logger);
+
+			if (doc.Root is null)
+				throw new InvalidOperationException ("Unix config file root is missing");
```
jonpryor, line 171: Should this include the path that couldn't be found?  If this exception "leaks" to the user, it's not particularly "actionable".

jonpryor: What's "doubly hilarious" is that this codepath should be unreachable, i.e. *it isn't possible for `doc.Root` to be `null`*:

https://github.com/xamarin/xamarin-android-tools/blob/7ef8ec51f4ca2ddd2f651b87b487c367ac833ceb/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkUnix.cs#L254-L256

`GetUnixConfigFile()` *ensures* that `doc.Root` isn't `null`!

And you've had to "duplicate" this check in multiple places.

Perhaps the "better" solution is to change the return type of `GetUnixConfigFile()` to `XElement` instead of `XDocument`, and have it be the "root" element?

jonpryor: and maybe update `SaveConfig()` to likewise take `XElement` instead of `XDocument`?

I'm not sure we *need* to use `XDocument` here…

PR #218, opened 2023-11-09
Bumps platforms-tools and cmdline-tools for d17.9

tondat:
- Bump platforms-tool to version 34.0.5
- Bump cmdline-tools to version 11.0
- Bump build-tool to version 34.0.0
- Bump NDK to version 26.1.10909125

- Ensures be in sync with Xamarin Android Feed for d17.9
- Enables the creation of pixel 6/7 on VisualStudio Device Manager tool

PR #219, opened 2023-11-15
Update the maximum NDK version to 26

grendello:
This tracks the current NDK LTS release.


dellis1972 (approved):

jonathanpeppers (approved):

pjcollins (approved):
LGTM this should lift the restriction on being able to use NDK 26, and we will also be looking at bumping the recommended/installed versions of all of our dependencies in https://github.com/xamarin/xamarin-android-tools/pull/218.

PR #220, opened 2023-11-17
[MSBuildReferences.projitems] Allow consumers to opt into `Microsoft.Build` package.

jpobst:
Projects that import `MSBuildReferences.projitems` receive a `<PackageReference>` to `Microsoft.Build`.  However, many consumers are `netstandard2.0` and this package does not have that framework, resulting in the following warnings:

```
Warning NU1701 Package 'Microsoft.Build 17.3.2' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.
Warning NU1701 Package 'Microsoft.IO.Redist 6.0.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.	
```

Most consumers do not actually need this package, so we can make it optional to prevent the warnings and extra dependencies in those projects.

Initially, this is done by making the package reference opt-in, with:
```xml
<_IncludeMicrosoftBuildPackage>true</_IncludeMicrosoftBuildPackage>
```

However this will be a "breaking" change to an unknown amount of consumers.  If we would like to avoid this, we can invert the change to allow projects to "opt-out" instead.

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/MSBuildReferences.projitems
+++ b/src/Microsoft.Android.Build.BaseTasks/MSBuildReferences.projitems
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build"                Version="$(MSBuildPackageReferenceVersion)" />
+    <PackageReference Include="Microsoft.Build"                Version="$(MSBuildPackageReferenceVersion)" Condition=" '$(_IncludeMicrosoftBuildPackage)' == 'true' " />
     <PackageReference Include="Microsoft.Build.Framework"      Version="$(MSBuildPackageReferenceVersion)" />
     <PackageReference Include="Microsoft.Build.Tasks.Core"     Version="$(MSBuildPackageReferenceVersion)" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MSBuildPackageReferenceVersion)" />


src/Microsoft.Android.Build.BaseTasks/MSBuildReferences.projitems
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build"                Version="$(MSBuildPackageReferenceVersion)" />
+    <PackageReference Include="Microsoft.Build"                Version="$(MSBuildPackageReferenceVersion)" Condition=" '$(_IncludeMicrosoftBuildPackage)' == 'true' " />
```
dellis1972, line 13: Should this be applied to all Microsoft.Build.* PackageReferences?

jpobst: We could add different properties for each `PackageReference` if we have a use case for them.  The use case we are currently looking at is we want to add a reference to all of the `Microsoft.Build.*` packages _except_ this one.

jonpryor:
After some discussion, my preference is to not introduce `$(_IncludeMicrosoftBuildPackage)` and instead make the `Microsoft.Build` package reference conditional on `$(TargetFramework)` *not* being `netstandard`.  This would likewise fix the NU1701 warnings, as `Microsoft.Build` would no longer be includerd in the builds of .NETStandard projects.

The one "exception" to this is `Xamarin.Android.Build.Tasks.csproj`, which still needs the project reference.  We would just need to add an "explicit" `<ProjectReference Include="Microsoft.Build" />` to `Xamarin.Android.Build.Tasks.csproj`, which will also emit the NU1701 warning, but that's unavoidable.

jpobst:
> After some discussion, my preference is to not introduce `$(_IncludeMicrosoftBuildPackage)` and instead make the `Microsoft.Build` package reference conditional on `$(TargetFramework)` *not* being `netstandard`.

After thinking some more about this, I think the original `$(_IncludeMicrosoftBuildPackage)` idea makes more sense.  It allows us to key this `<PackageReference>` on whether the consuming project needs the reference or not, rather than the consuming project's target framework.

The property allows us cleaner understanding of the intent rather than a hidden side-effect of target framework.

It's also cleaner to add the package to a project targeting both `netstandard` and something else:
https://github.com/xamarin/xamarin-android-tools/blob/08a69900df23e9f95057f3cacae72bd40d640a27/src/Microsoft.Android.Build.BaseTasks/Microsoft.Android.Build.BaseTasks.csproj#L6

jpobst:
Test XA bump: https://github.com/xamarin/xamarin-android/pull/8545

Fixes 226 reported warnings (390 -> 164).

PR #227, opened 2024-02-07
Only conditionally include <uses-sdk /> in the AndroidManifest.xml wh…

jstedfast:
…en saving

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1874249

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidAppManifest.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidAppManifest.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 using System.Xml;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


tondat (approved):

jonathanpeppers (commented):
So the design for these should be that:

* `$(TargetPlatformVersion)` is `<uses-sdk android:targetSdkVersion`
* `$(SupportedOSPlatformVersion)` is `<uses-sdk android:minSdkVersion`

`$(SupportedOSPlatformVersion)` is the only one that is possible to be blank, but it's in the project template.

Details here: https://learn.microsoft.com/en-us/dotnet/maui/migration/android-projects?view=net-maui-8.0#changes-to-androidmanifestxml

In what case are you able to make both of these blank?

jstedfast:
@jonathanpeppers net-android projects apparently don't build if AndroidManifest.xml has `<uses-sdk />`

What this patch does is remove the `uses-sdk` XML node if and only if it has no attributes.

Without this change, the `<uses-sdk />` XML element keeps getting added (back) to the AndroidManifest.xml file whenever the user edits that file in the UI editor in VS.

In other words, there's no way for the user to remove it unless they edit it in a text editor and manually remove it.

But if they go back into the UI editor ever in the future to change something, it will get added right back.

This patch fixes that scenario.

jonathanpeppers:
Can we add a unit test in here?

https://github.com/xamarin/xamarin-android-tools/blob/main/tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidAppManifestTests.cs

jstedfast:
@jonathanpeppers I just added some tests

jonathanpeppers (approved):

PR #230, opened 2024-04-10
Check for ANDROID_HOME Sdk location on Windows.

dellis1972:
For some reason we never check for the ANDROID_HOME environment variable on Windows. This might cause
issues with people using VSCode since we probably
will be using environment variables as a backup when using that IDE.

So lets check those last in the list of paths. This way we will pick up all the other defaults and backups before trying the environment variable.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
@@ -108,6 +108,10 @@ namespace Xamarin.Android.Tools
 			foreach (var basePath in paths)
 				if (Directory.Exists (basePath))
 					yield return basePath;
+
+			// check for environment variables last.
+			foreach (string dir in GetSdkFromEnvironmentVariables ())
+				yield return dir;
 		}
 
 		protected override IEnumerable<string> GetAllAvailableAndroidNdks ()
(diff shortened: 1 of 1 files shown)


dellis1972:
I added this becuse when writing the docs, I found that from the command line on windows you either have to 

1. Provide the AndroidSdkDirectory argument via `-p:AndroidSdkDirectory`.
2. do `set AndroidSdkDirectory=c:\foo`

`ANDROID_HOME` or `ANDROID_SDK_ROOT` just didn't work, they are ignored.

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
```diff
@@ -108,6 +108,10 @@ protected override IEnumerable<string> GetAllAvailableAndroidSdks ()
 			foreach (var basePath in paths)
 				if (Directory.Exists (basePath))
 					yield return basePath;
+
+			// check for environment variables last.
+			foreach (string dir in GetSdkFromEnvironmentVariables ())
+				yield return dir;
```
jonathanpeppers, line 114: Does this actually need to check "next to last"? Where it would check env vars before the hardcoded paths?

I'm thinking when `%localappdata%\Android\android-sdk` exists, and someone sets `%ANDROID_HOME%` it wouldn't work.

dellis1972: I'm not sure, it wouldn't pick it up at all with the current code, so I wasn't sure if I should change the existing behaviour (e.g it would pick up and use `%localappdata%\Android\android-sdk` rather than `ANDROID_HOME`.

jonathanpeppers (approved):

src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkWindows.cs
dellis1972: I moved this up a bit after the registry checks as per the chat on the meeting yesterday

PR #233, opened 2024-05-21
Update OneLocBuildToken

MSylvia:
Update the token in the Localization step to one backed by a managed identity


MSylvia:
This one might need more validation.

MSylvia:
The job is already using the correct pool. Should be fine to merge.

PR #235, opened 2024-05-27
Change class to public to use it on xamarinvs

thaystg:
Used in this PR: https://github.com/xamarin/XamarinVS/pull/14525


jonpryor:
I don't like this, as it makes private implementation details public.  This assembly is cross-platform, and I do not like the idea of having a public Windows-specific type.  Given that it's "just a bunch of P/Invokes", I'd feel better with you just *copying* the file wherever you need it.

I also don't see any *use* of `RegistryEx` within https://github.com/xamarin/XamarinVS/pull/14525, which just confuses me further.

PR #237, opened 2024-06-11
[BaseTasks] Import Xamarin.Build.AsyncTask

pjcollins:
We've decided to move https://github.com/xamarin/Xamarin.Build.AsyncTask
into this repository as part of dotnet org migration efforts.

Source and test content from https://github.com/xamarin/Xamarin.Build.AsyncTask/commit/db4ce14dac
has been added to Microsoft.Android.Build.BaseTasks.

The `AndroidAsyncTask` and `AsyncTask` classes have been merged, 
as we don't need both levels of abstraction.

`Microsoft.Android.Build.BaseTasks-Tests` has been updated to target
net8.0 as I noticed the tests were no longer running:

    /Users/runner/hostedtoolcache/dotnet/dotnet test /Users/runner/work/1/s/bin/TestDebug-net6.0/Microsoft.Android.Build.BaseTasks-Tests.dll --logger trx --results-directory /Users/runner/work/_temp
    Starting test execution, please wait...
    A total of 1 test files matched the specified pattern.
    No test is available in /Users/runner/work/1/s/bin/TestDebug-net6.0/Microsoft.Android.Build.BaseTasks-Tests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.


src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+using System.Linq;
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Framework;
+using System.Threading;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Android.Build.Tasks
+{
+    /// <summary>
```
jonpryor, line 15: Since we're copying the file, should we update it to use tabs for indentation?  That would be consistent with other files in this repo.

pjcollins:
Other testing PRs:

https://github.com/xamarin/androidtools/pull/443
https://github.com/xamarin/android-sdk-installer/pull/966
https://github.com/xamarin/monodroid/pull/1497
https://github.com/xamarin/xamarin-android/pull/9017

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+		/// </summary>
+		protected string WorkingDirectory { get; private set; }
+
+		[Obsolete("Do not use the Log.LogXXXX from within your Async task as it will Lock the Visual Studio UI. Use the this.LogXXXX methods instead.")]
+		private new TaskLoggingHelper Log => base.Log;
+
+		/// <summary>
+		/// Initializes the task.
+		/// </summary>
+		public AsyncTask()
```
jonpryor, line 60: …and otherwise reformat the code to be consistent with this repo: space before `(`, `{` on the same when appropriate (`if`, `foreach`, etc.), etc.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Android.Build.Tasks
+{
+	/// <summary>
+	/// Base class for tasks that need long-running cancellable asynchronous tasks 
+	/// that don't block the UI thread in the IDE.
+	/// </summary>
+	public class AsyncTask : Task, ICancelableTask
```
jonpryor, line 19: Since this is new API -- the namespace changed -- we should revisit some API design decisions:

 1. This class isn't *really* usable by itself; nobody will "reasonably" do `new AsyncTask()`.  As such, should we make this type `abstract` and make the constructor `protected`?
 2. Should we "flatten" `AsyncTask` and `AndroidAsyncTask` / "Task-ify `AsyncTask`, by way of moving `RunTaskAsync()` and `RunTask()` to this type?  (Leave `TaskPrefix` in `AndroidAsyncTask`.)
 3. Variation on (2): it's not at all obvious to subclasses what they should *do*.  It declares no new `virtual` or `abstract` methods, meaning `Execute()` is the only obvious method to override.  (2) could help solve that.

Is there any other API cleanup we should consider?

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+			}
+
+			var data = new TelemetryEventArgs() {
+					EventName = eventName,
+					Properties = properties
+			};
+			EnqueueMessage(telemetryMessageQueue, data, telemetryDataAvailable);
+		}
+
+		public void LogMessage(string message) => LogMessage(message, importance: MessageImportance.Normal);
```
jonpryor, line 120: This is probably a "me" thing, but this line just looks "too wide"; I'd like it more if it were line wrapped:

```csharp
// Option 1
public void LogMessage (string message) =>
    LogMessage (message, importance: MessageImportance.Normal);

// Option 2
public void LogMessage (string message)
    => LogMessage (message, importance: MessageImportance.Normal);
```

I'm not sure I have a preference between Options (1) or (2), just that I don't like "really wide" lines.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+
+		public void LogCodedError(string code, string file, int lineNumber, string message, params object[] messageArgs) => LogCodedError(code: code, message: string.Format(message, messageArgs), file: file, lineNumber: lineNumber);
+
+		public void LogCodedError(string code, string message, string file, int lineNumber)
+		{
+			if (uiThreadId == Thread.CurrentThread.ManagedThreadId)
+			{
+#pragma warning disable 618
+				Log.LogError(
+					subcategory: null,
```
jonpryor, line 202: Indentation is inconsistent.  *My* preference is "parameters indented *two* tab stops*, which is done ~15 lines below!  *Here* parameters are indented *one* tab stop, resulting in "odd" alignment wherein *this* invocation, nested in an `if`, has the same indentation as the subsequent invocation, which *isn't* nested in an `if`!

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
dellis1972: I'm in favour of making it abstract and keeping the existing tooling. Because a custom task generally involves subclassing and writting code I would not expect anyone to try using this class directly. 

When this was a generic class to be used anywhere it made sense to have our own `AndroidAsyncTask` to make things easier for us. As that is no longer required we can probably folder the AndroidAsyncTask into this one. The only down side on that is the number of changes that will be needed when we bump.

pjcollins: I like the idea of moving some things into `AsyncTask`, while keeping `AndroidAsyncTask` around to reduce the number of changes required. Though, we could also remove `AndroidAsyncTask`, and merge everything into a single `AsyncTask` (or continue using the `AndroidAsyncTask` name) if that would be preferable?

pjcollins: I kept these as two separate classes for now, while moving most functionality into the now abstract AsyncTask.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,431 @@
...
+				return;
+#pragma warning restore 618
+			}
+
+			StackFrame exceptionFrame = null;
+			try
+			{
+				exceptionFrame = new StackTrace(exception, true)?.GetFrames()?.FirstOrDefault();
+			}
+			catch { }
```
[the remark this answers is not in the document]
pjcollins: Looks like this will only throw if exception is null, so I added a null check and removed the try/catch https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stacktrace.-ctor?view=net-8.0#system-diagnostics-stacktrace-ctor(system-exception-system-boolean)

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -0,0 +1,461 @@
...
+		}
+
+		public void Cancel ()
+			=> taskCancelled.Set ();
+
+		protected void Complete (System.Threading.Tasks.Task task)
+		{
+			if (task.Exception != null) {
+				var ex = task.Exception.GetBaseException ();
+				LogError (ex.Message + Environment.NewLine + ex.StackTrace);
```
jonpryor, line 75: @dellis1972: any recollection of why we logged the error this way instead of just logging `ex.ToString()`?

@dellis1972, @pjcollins: with the realization that this next comment goes well into "the perfect is the enemy of the good"… Should we move `AndroidAsyncTask.TaskPrefix` into this type so that we can use `LogCodedError()` keyed off of `TaskPrefix` here and elsewhere?

pjcollins: > @dellis1972, @pjcollins: with the realization that this next comment goes well into "the perfect is the enemy of the good"… Should we move AndroidAsyncTask.TaskPrefix into this type so that we can use LogCodedError() keyed off of TaskPrefix here and elsewhere?

I think this goes back to the question above -- it probably does make most sense to merge the two classes and deal with the renaming/breakage that comes with that.

dellis1972: We can merge them if we want, I was just trying to make the bumps a bit easier. But if its better in the long run then go for it 👍

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
pjcollins: `AndroidAsyncTask` and `AsyncTask` have been merged into a single `AsyncTask`.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
pjcollins: Merged the two, and updated this Exception logging function to use LogUnhandledException which will output an error code using TaskPrefix.

pjcollins:
@dellis1972 @jonathanpeppers I think this should be ready for further review and hopefully good to go now. I just bumped `Microsoft.Android.Build.BaseTasks-Tests` to .NET 8 as I noticed those tests weren't actually running in main:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9730201&view=logs&j=3b059b26-55fe-5273-567a-5b3fa6c051e7&t=73552669-1d50-5779-0733-0594204b4f2d&l=16

The downstream PRs are all looking good so far against these latest changes as well.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -157,40 +157,6 @@ public void LogError (string message)
 		public void LogError (string message, params object [] messageArgs)
 			=> LogCodedError (code: null, message: string.Format (message, messageArgs));
 
-		public void LogErrorFromException (Exception exception)
```
dellis1972: This API might be needed at some point by classes deriving from AsyncTask. Its part of the public API.

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
```diff
@@ -75,7 +75,7 @@ protected void Complete (System.Threading.Tasks.Task task)
 		{
 			if (task.Exception != null) {
 				var ex = task.Exception.GetBaseException ();
-				LogError (ex.Message + Environment.NewLine + ex.StackTrace);
+				this.LogUnhandledException (TaskPrefix, ex);
```
dellis1972, line 78: This may well lock up the UI thread if it is called from a Task. This is one of the reasons we have custom versions of all the logging.

dellis1972: The lock up might not show up in our uni tests as they are using a Mock Engine.

pjcollins: I think this one is thread safe, as we have this [AsyncTask specific extension](https://github.com/dotnet/android-tools/blob/1ea4e35eaf3fe09b9ca8adf0a29b4d9988146338/src/Microsoft.Android.Build.BaseTasks/UnhandledExceptionLogger.cs#L16)? I think this should ultimately use https://github.com/xamarin/Xamarin.Build.AsyncTask/blob/db4ce14dacfef47435c238b1b681c124e60ea1a0/Xamarin.Build.AsyncTask/AsyncTask.cs#L156?

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
pjcollins: This is only defined in the [android-sdk-installer copy of AsyncTask](https://github.com/xamarin/android-sdk-installer/blob/7455826102c50db08e34b044aa5348b8b4d459fe/Xamarin.Installer.Build.Tasks/Tasks/AsyncTask.cs#L167), which I was planning to remove after this addition is merged and bumped. I think we should be able to use the [AsyncTask specific extension](https://github.com/dotnet/android-tools/blob/1ea4e35eaf3fe09b9ca8adf0a29b4d9988146338/src/Microsoft.Android.Build.BaseTasks/UnhandledExceptionLogger.cs#L16) instead, and remove this one specific call site if that seems fair?

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
dellis1972: ok cool. I was not aware of that extension method. Should we move the method ?

src/Microsoft.Android.Build.BaseTasks/AsyncTask.cs
dellis1972: That might be ok. I just want to make sure we are covered because if people use `Log.XXX` based methods in a Task it will lock up.

jonathanpeppers (approved):
LGTM 👍 The main fallout will be bumping a lot of things everywhere!

PR #240, opened 2024-08-06
Be explicit about FileShare mode when using FileStream.

dellis1972:
Context https://dev.azure.com/devdiv/DevDiv/_workitems/edit/2055618

We know that by default `FileShare` will be `Read` when using a `FileStream`. However lets be totally clear on what we want from this point on, so if there is a problem we know its not us.


jonathanpeppers (approved):
Sometimes I use `File.OpenRead()` instead of `FileStream` ctor, but I think either of these is fine.

PR #241, opened 2024-08-16
Bump to dotnet/msbuild@10fbfbf2

pjcollins:
Changes: https://github.com/dotnet/msbuild/compare/v17.9.5...v17.10.4

Updates $(MSBuildPackageReferenceVersion) to the latest VS 2022 LTS.

This new package version brings in System.Security.Cryptography.Xml
v8.0.0 as a transitive dependency, and as such the explicit v7.0.1
package reference should no longer be necessary.

An explicit package reference has been added for System.Formats.Asn1
v8.0.1 (another transitive dependency of Microsoft.Build) to address
https://github.com/advisories/GHSA-447r-wph3-92pm.

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/MSBuildReferences.projitems
+++ b/src/Microsoft.Android.Build.BaseTasks/MSBuildReferences.projitems
@@ -4,8 +4,7 @@
 <Project>
   <!--Import this file in projects needing to reference Microsoft.Build.*.dll -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


pjcollins:
Testing further with https://github.com/dotnet/android/pull/9216

jonathanpeppers (approved):
When I've updated `$(MSBuildPackageReferenceVersion)` in the past, it has been a pain to bump everything. I think things break until I had the change everywhere: android-sdk-installer, monodroid, dotnet/android. Just fyi.

I think we _have_ to update the occasionally, it's just painful.

dellis1972 (approved):

PR #242, opened 2024-09-04
[Microsoft.Android.Build.BaseTasks] automatically report locked files

jonathanpeppers:
Context: https://github.com/dotnet/msbuild/commit/5f27e4d861a19ecff094161ff203a8a5924bd3d2

Since ~2017, MSBuild has added useful context to the `<Copy/>` task that reports *which* process locks a certain file on Windows. This logic has slowly been added to more places in MSBuild.

We should add this logic to `Microsoft.Android.Build.BaseTasks` as well!

* If an `AndroidTask` gets an `IOException`

* And we can parse a file path between the `''` characters

* And the file exists

Report an error message such as:

    error XAMYT7024: The file "D:\temp\tmphkqpda.tmp" is locked by: "MSBuild (22040)"
    System.IO.IOException: The process cannot access the file 'D:\temp\tmphkqpda.tmp' because it is being used by another process.
       at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile (String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
    ... rest of stacktrace

This requires us to bring over two files from MSBuild:

* `LockCheck.cs`
* `NativeMethodsShared.cs`

I only made a couple small changes to `LockCheck.cs`:

* Format the entire file with `dotnet-format`

* For localization, use `Resources.LockCheck_FileLocked` directly

* Change visibility of various methods

* Only `LockCheck.GetLockedFileMessage()` is public


src/Microsoft.Android.Build.BaseTasks/Microsoft.Android.Build.BaseTasks.csproj
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <RootNamespace>Microsoft.Android.Build.Tasks</RootNamespace>
+    <LangVersion>latest</LangVersion>
```
jonathanpeppers, line 8: I added this, otherwise the `netstandard2.0` target couldn't build `LockFile.cs`.

jonpryor:
It occurs to me, belatedly, that `LockCheck` is *public*: https://github.com/dotnet/msbuild/blob/7cf66090a764f0f239671e4877255efe7ba91155/src/Utilities/LockCheck.cs#L22

And *documented*: [`LockCheck.GetLockedFileMessage(string)`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.lockcheck.getlockedfilemessage?view=msbuild-17-netcore)

It's in "Microsoft.Build.Utilities.Core v17.11.4".

Should we use it?

The downside is that we can't easily customize the message for our own use, e.g. "elsewhere" I had mentioned that it might be handy if `GetLockingProcessInfos()` included [`Process.GetCurrentProcess()`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.getcurrentprocess?view=net-8.0), as the whole reason we're getting a "file share exception" is because the current process is *also* trying to use the file, so *maybe* that would be handy to know?

dellis1972 (approved):

PR #245, opened 2024-10-14
Add a retry attempt when copying files

dellis1972:
Context https://github.com/dotnet/android/issues/9133.

We get collisions between the DTB (or AV) and our main build. This can result in the following error

```
Error (active)	XALNS7019	System.UnauthorizedAccessException: Access to the path 'D:\Projects\MauiApp2\obj\Debug\net9.0-android\android\assets\armeabi-v7a\MauiApp2.dll' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at Microsoft.Android.Build.Tasks.Files.CopyIfChanged(String source, String destination) in /Users/runner/work/1/s/xamarin-android/external/xamarin-android-tools/src/Microsoft.Android.Build.BaseTasks/Files.cs:line 125
   at Xamarin.Android.Tasks.MonoAndroidHelper.CopyAssemblyAndSymbols(String source, String destination) in /Users/runner/work/1/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs:line 344
   at Xamarin.Android.Tasks.LinkAssembliesNoShrink.CopyIfChanged(ITaskItem source, ITaskItem destination) in /Users/runner/work/1/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs:line 161
   at Xamarin.Android.Tasks.LinkAssembliesNoShrink.RunTask() in /Users/runner/work/1/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs:line 76
   at Microsoft.Android.Build.Tasks.AndroidTask.Execute() in /Users/runner/work/1/s/xamarin-android/external/xamarin-android-tools/src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs:line 25	MauiApp2 (net9.0-android)	C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\34.99.0-preview.6.340\tools\Xamarin.Android.Common.targets	1407
```

If we look at the MSBuild [Copy](https://github.com/dotnet/msbuild/blob/main/src/Tasks/Copy.cs#L897) task we see that it has a retry system in the cases of `UnauthorizedAccessException` or `IOException` when the code is `ACCESS_DENIED` or `ERROR_SHARING_VIOLATION`. So lets duplicate that kind of logic in our `Copy*IfChanged` helper methods. This should give our builds a bit more resiliency to these kinds of issues.

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/Files.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/Files.cs
@@ -160,8 +160,7 @@ namespace Microsoft.Android.Build.Tasks
 			int retryCount = 0;
 			int attempts = GetFileWriteRetryAttempts ();
 			int delay = GetFileWriteRetryDelay ();
-			while (retryCount < attempts) {
-				Console.WriteLine ($"{nameof(CopyIfChanged)}: Copy Attempt {retryCount}.");
+			while (retryCount <= attempts) {
 				try {
 					return CopyIfChangedRetry (source, destination);
 				} catch (Exception e) {
@@ -170,7 +169,6 @@ namespace Microsoft.Android.Build.Tasks
 						case IOException:
 							int code = Marshal.GetHRForException (e);
 							if ((code != ERROR_ACCESS_DENIED && code != ERROR_SHARING_VIOLATION) || retryCount == attempts) {
-								Console.WriteLine ($"{nameof(CopyIfChanged)}: Throwing {e}\ncode:{code}\nretryCount:{retryCount}.");
 								throw;
 							};
 							break;
@@ -179,7 +177,6 @@ namespace Microsoft.Android.Build.Tasks
 					} 
 				}
 				retryCount++;
-				Console.WriteLine ($"{nameof(CopyIfChanged)}: Copy Sleeping {delay}.");
 				Thread.Sleep (delay);
 			}
 			return false;
@@ -236,7 +233,7 @@ namespace Microsoft.Android.Build.Tasks
 			int retryCount = 0;
 			int attempts = GetFileWriteRetryAttempts ();
 			int delay = GetFileWriteRetryDelay ();
-			while (retryCount < attempts) {
+			while (retryCount <= attempts) {
 				try {
 					return CopyIfStreamChangedRetry (stream, destination);
 				} catch (Exception e) {
--- a/src/Microsoft.Android.Build.BaseTasks/Files.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/Files.cs
@@ -9,6 +9,10 @@ using System.Security.Cryptography;
 using System.Text;
 using Xamarin.Tools.Zip;
 using Microsoft.Build.Utilities;
+using System.Threading;
+using System.Reflection.Metadata;
+using System.Runtime.InteropServices;
+using System.Collections;
 
 namespace Microsoft.Android.Build.Tasks
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Android.Build.BaseTasks/Files.cs
```diff
@@ -28,6 +38,33 @@ public static class Files
 		public static readonly Encoding UTF8withoutBOM = new UTF8Encoding (encoderShouldEmitUTF8Identifier: false);
 		readonly static byte[] Utf8Preamble = Encoding.UTF8.GetPreamble ();
 
+		/// <summary>
+		/// Checks for the environment variable DOTNET_ANDROID_FILE_WRITE_RETRY_ATTEMPTS to
+		/// see if a custom value for the number of times to retry writing a file has been 
+		/// set.
+		/// </summary>
+		/// <returns>The value of DOTNET_ANDROID_FILE_WRITE_RETRY_ATTEMPTS or the default of DEFAULT_FILE_WRITE_RETRY_ATTEMPTS</returns>
+		public static int GetFileWriteRetryAttempts ()
```
dellis1972, line 47: I made these public so we can use them from XA.

src/Microsoft.Android.Build.BaseTasks/Files.cs
```diff
@@ -28,6 +38,33 @@ public static class Files
...
+		{
+			var retryVariable = Environment.GetEnvironmentVariable ("DOTNET_ANDROID_FILE_WRITE_RETRY_ATTEMPTS");
+			if (!string.IsNullOrEmpty (retryVariable) && int.TryParse (retryVariable, out int retry))
```
jonpryor, line 50: The value should be cached on first access.  Unless you have a unit test setting this variable, I don't see the need to have the performance overhead of checking + parsing this value all the time.

src/Microsoft.Android.Build.BaseTasks/Files.cs
```diff
@@ -157,6 +229,33 @@ public static bool CopyIfBytesChanged (byte[] bytes, string destination)
...
+			int delay = GetFileWriteRetryDelay ();
+			while (retryCount <= attempts) {
+				try {
+					return CopyIfStreamChangedOnce (stream, destination);
+				} catch (Exception e) {
+					switch (e) {
+						case UnauthorizedAccessException:
+						case IOException:
+							int code = Marshal.GetHRForException (e);
+							if ((code != ERROR_ACCESS_DENIED && code != ERROR_SHARING_VIOLATION) || retryCount == attempts) {
```
jonathanpeppers, line 244: For general safety, should we do this to prevent an infinite loop:
```suggestion
							if ((code != ERROR_ACCESS_DENIED && code != ERROR_SHARING_VIOLATION) || retryCount >= attempts) {
```

src/Microsoft.Android.Build.BaseTasks/Files.cs
jonathanpeppers: Do we have an other environment variables like this? I wonder if they should go in a class like `KnownEnvironment.cs`.

jonathanpeppers (approved):
Overall, this looks good to me, just had a couple comments.

src/Microsoft.Android.Build.BaseTasks/Files.cs
dellis1972: We don't afaik.

jonpryor:
@jonathanpeppers [asked](https://github.com/dotnet/android-tools/pull/245#discussion_r1810979615):

> Do we have an other environment variables like this?

Yes?  `git grep GetEnviron` shows a numb er of environment variables that we use for non-test purposes, including:

  * `JI_JAVA_HOME`: `JAVA_HOME` "override" for Java.Interop
  * `JAVA_HOME`: JDK install directory
  * `XDG_CACHE_HOME`: Cache directory
  * `XAMARIN_ANDROID_REGKEY`: Registry key which contains the Android SDK directory on Windows
  * `ANDROID_HOME`, `ANDROID_SDK_ROOT`: Android SDK directory
  * …and others…

I'm not sure what a `KnownEnvironment.cs` would necessarily buy us.

PR #246, opened 2024-11-11
[Xamarin.Android.Tools.AndroidSdk] Filter out invalid JDKs

jonpryor:
@jonpryor's developer machines are a "weird", with a macOS arm64 box containing the directories:

  * `$HOME/android-toolchain/jdk-17`
  * `$HOME/android-toolchain/jdk-17.x64`

Meanwhile, `XAPrepareJdkLocations.GetXAPrepareJdks()` looks for all directories matching the glob `$HOME/android-toolchains/jdk*`, and thus finds both.

The one that dotnet/java-interop winds up using by default is the the `jdk-17.x64` directory (😱), which cannot be loaded into an arm64 process.  This results in build failures.

Instead of doing the "obvious" of removing `jdk.x64` or renaming it so it doesn't match the glob `jdk*`, update
`JdkInfo.GetKnownSystemJdkInfos()` and `JdkInfo.GetSupportedJdkInfos()` to filter out JDKs that cannot be loaded into the current process. This nicely filters out `jdk-17.x64`, allowing dotnet/java-interop to build successfully.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/JdkInfo.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -333,6 +334,7 @@ namespace Xamarin.Android.Tools
 				.Concat (GetPathEnvironmentJdks (logger))
 				.Concat (GetLibexecJdks (logger))
 				.Concat (GetJavaAlternativesJdks (logger))
+				.Where (v => JdkRunsOnHost (v, logger))
 				;
 		}
 
@@ -344,7 +346,23 @@ namespace Xamarin.Android.Tools
 			return MicrosoftOpenJdkLocations.GetMicrosoftOpenJdks (logger)
 				.Concat (EclipseAdoptiumJdkLocations.GetEclipseAdoptiumJdks (logger))
 				.Concat (MicrosoftDistJdkLocations.GetMicrosoftDistJdks (logger))
+				.Where (v => JdkRunsOnHost (v, logger))
 				;
+		}
+
+		static bool JdkRunsOnHost (JdkInfo jdk, Action<TraceLevel, string> logger)
+		{
+			if (OS.IsMac) {
+				var cputype = RuntimeInformation.OSArchitecture;
+				if (jdk.ReleaseProperties.TryGetValue ("OS_ARCH", out var arch)) {
+					return (cputype, arch) switch {
+						(Architecture.Arm64,    "aarch64")  => true,
+						(Architecture.X64,      "x86_64")   => true,
+						_ => false,
+					};
+				}
+			}
+			return true;
 		}
 
 		internal static JdkInfo? TryGetJdkInfo (string path, Action<TraceLevel, string> logger, string locator)


jpobst:
I assume this is for the "desktop Java binding" sample we have in `java.interop` which tries to load Java into an existing process.

Is this filter something we want for "normal" Android?  That is, if I have an ARM Mac that only has x64 Java on it, shouldn't that work via Rosetta?

jonpryor:
@jpobst [wrote](https://github.com/dotnet/android-tools/pull/246#issuecomment-2468774683):

> However, is this filter something we would want for "normal" .NET for Android?

*At present*, nothing in .NET for Android loads a Desktop JVM, so this will not impact current customers.  (*I want* .NET for Android to actually do that, as part of `jnimarshalmethod-gen`, but I rather doubt it'll be Good Enough™ anytime soon to actually matter here.)

> That is, if I have an arm64 Mac that only has x64 Java on it, shouldn't that work via Rosetta?

No and yes. An arm64 process cannot load x64 binaries, even with Rosetta.  If you want to use Rosetta, you'd have to install an x64 `dotnet`, use Rosetta to run the x64 `dotnet`, and *then* you could load an x64 Java into it.

Though this does suggest that I should instead use [`RuntimeInfo.ProcessArchitecture`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.processarchitecture?view=net-8.0#system-runtime-interopservices-runtimeinformation-processarchitecture) instead of `OSArchitecture`, so that we use the architecture of the current process, *not* the host OS, so that in a "Rosetta running `dotnet`" scenario we'll pick the right architecture.

> Maybe this needs to be done somewhere at a higher level where GetKnownSystemJdkInfos still returns all known system JDKs and a caller determines which one(s) are valid for its scenario?

This is an alternate solution, certainly.  I just figured that every caller that wanted to load a JVM would need to do similar logic, so doing it centrally would be preferred. (Though `JdkInfo` is *also* used to just find a `java`/`javac` to execute, which is a process boundary, and thus Rosetta/related emulation would be automatically used.

jpobst:
> Though JdkInfo is also used to just find a java/javac to execute, which is a process boundary, and thus Rosetta/related emulation would be automatically used.

This is the scenario I was thinking of, such as compiling JCWs.

jonpryor:
To verify Rosetta behavior, consider this `dotnet new console -n processor-architecture` program:

```csharp
using System.Runtime.InteropServices;

// See https://aka.ms/new-console-template for more information
Console.WriteLine($"       OSArchitecture: {RuntimeInformation.OSArchitecture}");
Console.WriteLine($"ProcessorArchitecture: {RuntimeInformation.ProcessArchitecture}");
```

Build:

```dotnetcli
dotnet build
```

Run with native arm64 `dotnet`:

```dotnetcli
% dotnet ./bin/Debug/net8.0/processor-architecture.dll
       OSArchitecture: Arm64
ProcessorArchitecture: Arm64
```

Run under x64 `dotnet` via Rosetta:

```dotnetcli
% $HOME/Downloads/dotnet-sdk-8.0.403-osx-x64/dotnet bin/Debug/net8.0/processor-architecture.dll
       OSArchitecture: Arm64
ProcessorArchitecture: X64
```

*Note*: to allow the x64 dotnet to run, I had to apply @lambdageek's [fix to the installation directory](https://github.com/dotnet/runtime/issues/95259#issuecomment-1828007858), otherwise it would crash.

jonpryor:
@jpobst [wrote](https://github.com/dotnet/android-tools/pull/246#issuecomment-2468821817):

> This is the scenario I was thinking of, such as compiling JCWs.

That's a fair point.  This logic should be move into dotnet/java-interop then. (It should *also* use `.ProcessorArchitecture` and not `.OSArchitecture`…)

PR #251, opened 2025-04-21
[build] Update to net9.0 and netstandard2.0

pjcollins:
Introduces a `$(DotNetTargetFramework)` property to control the target
framework version used for exe projects, and sets it to the latest
stable .NET version (`net9.0`).

Library and build task projects have been updated to only target
`netstandard2.0`, which is compatible with both .NET Framework and .NET
Core.

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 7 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.AndroidSdk.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
jonathanpeppers, line 4: I can't think of anything that would require the `net6.0` assembly produced by this project.

dellis1972 (approved):

PR #282, opened 2026-02-23
Add AvdManagerRunner for avdmanager CLI operations

rmarinho:
## Add AvdManagerRunner for avdmanager CLI operations

Adds `AvdManagerRunner` — a typed async wrapper around the Android `avdmanager` CLI tool for creating, listing, and deleting Android Virtual Devices.

Also adds `FindCmdlineTool` to `ProcessUtils` — discovers cmdline-tools binaries by checking the `latest` symlink first, then falling back to the highest versioned directory.

### API Surface

```csharp
public class AvdManagerRunner
{
    public AvdManagerRunner(string avdManagerPath,
        IDictionary<string, string>? environmentVariables = null,
        Action<TraceLevel, string>? logger = null);

    public Task<IReadOnlyList<AvdInfo>> ListAvdsAsync(CancellationToken cancellationToken = default);
    public Task<AvdInfo> GetOrCreateAvdAsync(string name, string systemImage,
        string? deviceProfile = null, bool force = false,
        CancellationToken cancellationToken = default);
    public Task DeleteAvdAsync(string name, CancellationToken cancellationToken = default);
}

public record AvdInfo(string Name, string? DeviceProfile = null, string? Path = null);
```

```csharp
// ProcessUtils (internal)
internal static string? FindCmdlineTool(string sdkPath, string toolName, string extension,
    Action<TraceLevel, string>? logger = null);
```

### Key Design Decisions

- **`AvdInfo` is a positional `record`** — immutable, value equality, minimal boilerplate.
- **Resolved-path constructor**: Caller resolves the full avdmanager path upfront (consistent with `AdbRunner`/`EmulatorRunner` pattern).
- **`Action<TraceLevel, string>? logger`**: Optional diagnostic logger on constructor (consistent with all other runners). Logs command execution, get-or-create shortcuts, idempotent deletes, and IOException diagnostics.
- **Get-or-create pattern**: `GetOrCreateAvdAsync` returns an existing AVD if one matches by name when `force=false`. Callers who need to overwrite orphaned AVDs pass `force: true` explicitly.
- **Idempotent delete**: `DeleteAvdAsync` succeeds silently when the AVD doesn't exist.
- **`FindCmdlineTool`**: Checks `latest` directory first (per SDK convention), then versioned directories sorted descending with stable preferred over prerelease. No legacy `tools/bin` fallback (different CLI args). Uses `Action<TraceLevel, string>? logger` for diagnostic messages (not `Debug.WriteLine`).

### Tests (25 total)

| Category | Count | What |
|----------|-------|------|
| Parse AVD list | 6 | Multiple AVDs, Windows newlines, empty, no AVDs, single missing device, IReadOnlyList return type |
| Constructor validation | 4 | Null/empty/whitespace throws, accepts env vars |
| GetOrCreateAvdAsync validation | 6 | Null/empty/whitespace for both name and systemImage |
| DeleteAvdAsync validation | 3 | Null/empty/whitespace name |
| FindCmdlineTool | 6 | Versioned dir, higher version, prerelease, latest-first, missing SDK, stable-over-prerelease |

Plus 2 CI-gated integration tests (skipped when no SDK installed).

### Review Feedback Addressed

| Reviewer | Issue | Fix |
|----------|-------|-----|
| @jonathanpeppers | `Debug.WriteLine` doesn't work for MSBuild tasks | Added `Action<TraceLevel, string>? logger` param to `FindCmdlineTool` |
| @jonathanpeppers | Check `latest` first, then versioned dirs | Reordered: `latest` → versioned dirs |
| @jonathanpeppers | Never use legacy `tools` folder | Removed entirely — returns null |
| @jonathanpeppers | Empty catch blocks around File.Exists | Removed (File.Exists never throws) |
| @jonathanpeppers | YAGNI: ValidateNotNullOrEmpty unused | Removed entirely |
| @jonathanpeppers | AvdInfo should be record | Converted to positional record |
| Bot | CreateAvdAsync get-or-create naming | Renamed to `GetOrCreateAvdAsync` |
| Bot | IOException catch missing logging | Added `logger?.Invoke` in catch |
| Review | GetAvdRootDirectory duplicated SDK logic | Removed — callers pass `force: true` for orphans |
| Review | FindCmdlineTool stable-over-prerelease tiebreak | Secondary sort: stable dirs preferred |

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Runners/AvdManagerRunner.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Runners/AvdManagerRunner.cs
@@ -68,11 +68,6 @@ public class AvdManagerRunner
 				logger?.Invoke (TraceLevel.Verbose, $"AVD '{name}' already exists, returning existing");
 				return existing;
 			}
-
-			// Detect orphaned AVD directory (folder exists without .ini registration).
-			var avdDir = Path.Combine (GetAvdRootDirectory (), $"{name}.avd");
-			if (Directory.Exists (avdDir))
-				force = true;
 		}
 
 		var args = new List<string> { "create", "avd", "-n", name, "-k", systemImage };
@@ -152,43 +147,5 @@ public class AvdManagerRunner
 		return avds;
 	}
 
-	/// <summary>
-	/// Resolves the AVD root directory, respecting ANDROID_AVD_HOME and ANDROID_USER_HOME.
-	/// Checks instance <see cref="environmentVariables"/> first, then falls back to process environment.
-	/// </summary>
-	string GetAvdRootDirectory ()
-	{
-		// ANDROID_AVD_HOME takes highest priority
-		if (TryGetEnvironmentVariable (EnvironmentVariableNames.AndroidAvdHome, out var avdHome))
-			return avdHome;
-
-		// ANDROID_USER_HOME/avd is the next option
-		if (TryGetEnvironmentVariable (EnvironmentVariableNames.AndroidUserHome, out var userHome))
-			return Path.Combine (userHome, "avd");
-
-		// Default: ~/.android/avd
-		return Path.Combine (
-			Environment.GetFolderPath (Environment.SpecialFolder.UserProfile),
-			".android", "avd");
-	}
-
-	/// <summary>
-	/// Looks up an environment variable, checking instance <see cref="environmentVariables"/> first,
-	/// then falling back to the process environment.
-	/// </summary>
-	bool TryGetEnvironmentVariable (string name, out string value)
-	{
-		if (environmentVariables is not null && environmentVariables.TryGetValue (name, out var dictValue) && dictValue is { Length: > 0 }) {
-			value = dictValue;
-			return true;
-		}
-		var envValue = Environment.GetEnvironmentVariable (name);
-		if (envValue is { Length: > 0 }) {
-			value = envValue;
-			return true;
-		}
-		value = string.Empty;
-		return false;
-	}
 }
 


src/Xamarin.Android.Tools.AndroidSdk/Runners/AvdManagerRunner.cs
```diff
@@ -0,0 +1,181 @@
...
+
+	public async Task<AvdInfo> CreateAvdAsync (string name, string systemImage, string? deviceProfile = null,
+		bool force = false, CancellationToken cancellationToken = default)
+	{
+		if (string.IsNullOrWhiteSpace (name))
+			throw new ArgumentException ("Value cannot be null or whitespace.", nameof (name));
+		if (string.IsNullOrWhiteSpace (systemImage))
+			throw new ArgumentException ("Value cannot be null or whitespace.", nameof (systemImage));
+
+		// Check if AVD already exists — return it instead of failing
```
jonathanpeppers: 🤖 **Naming** - When force is false and an AVD with the same name already exists, CreateAvdAsync silently returns the existing AVD (lines 54-59). This is a get-or-create pattern, but the method name implies it always creates.

Consider either:
- Renaming to GetOrCreateAvdAsync, or
- Adding a summary doc comment that explicitly describes the return-existing behavior so callers are not surprised.

The current design is sound for idempotent setup scenarios - just needs the name or docs to communicate it.

_Rule: Method names must reflect behavior (Postmortem #36)_

jonathanpeppers (commented):
## 🤖 AI Review Summary

**Verdict**: LGTM with minor suggestions

Found **3 suggestions** - nothing blocking.

- Naming: CreateAvdAsync silently returns existing AVDs (get-or-create pattern) - consider documenting (AvdManagerRunner.cs:54)
- Error handling: Empty catch (IOException) in CreateAvdAsync stdin callback (AvdManagerRunner.cs:84)
- Code organization: Two lines missing tab indentation in tests (AvdManagerRunnerTests.cs:147, 279)

Positive callouts:
- Mirrors AdbRunner structure cleanly - constructor pattern, ProcessUtils usage, environment variable handling all consistent.
- AvdInfo as positional record is excellent - value equality, immutability, and matches SdkPackage/SdkLicense precedent.
- FindCmdlineTool version resolution (stable-over-prerelease tiebreaking) is well thought out.
- Comprehensive test coverage including Windows newlines, pre-release version dirs, orphaned AVD detection, and edge-case parsing.
- IReadOnlyList return types, CancellationToken properly propagated to all downstream calls, exit codes checked consistently.
- Good use of EnvironmentVariableNames constants throughout.

Note on PR description: The body states FindCmdlineTool prefers highest versioned directory, falls back to latest then legacy tools/bin but the code prefers latest first (correct behavior per SDK convention) and has no tools/bin fallback. Worth correcting so the description matches the implementation.

---
_Review generated by android-tools-reviewer._

PR #293, opened 2026-03-10
Fix CS8604 nullable warning in AndroidVersion.AlternateIds

rmarinho:
## Summary

Fix the CS8604 compiler warning in \AndroidVersion.AlternateIds\ setter where \HashSet<string>.UnionWith()\ was called with a potentially null \string[]?\ value.

## Changes

- **\AndroidVersion.cs\**: Guard \UnionWith(value)\ with \if (value is { Length: > 0 })\ pattern match — skips the call when value is null or empty.

## Build

- 0 warnings, 0 errors (both \
etstandard2.0\ and \
et10.0\)
- 197/208 tests pass (1 pre-existing env-dependent failure, 10 skipped)

Fixes the warning seen in [build 1323580](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1323580&view=results).


jonathanpeppers (commented):
Can we undo all the whitespace changes? That makes it difficult to review git history or git blame.

PR #299, opened 2026-03-11
[skills] Incorporate review feedback from dotnet/android#10918

jonathanpeppers:
Fix submit_review.cs issues found during porting:
- Replace GetString()! with ?? "" to avoid null-forgiving operator
- Add null check for Process.Start() instead of !
- Narrow bare catch blocks to catch (JsonException)
- Restructure success output so catch block isn't misleading

Merge C#-focused review rules back into review-rules.md:
- Async: thread safety of shared state
- Error handling: actionable exception details, challenge exception swallowing
- Code org: no commented-out code, early returns, skip default field init
- Performance: pre-allocate collections, avoid closures, cheap checks first, O(n²)
- Patterns: method names reflect behavior, why-not-what comments, track TODOs
- YAGNI: git commit --amend, commit message quality, typos, filler words
- New section 13: Testing (regression tests, specific assertions, deterministic data, edge cases)


jonathanpeppers (commented):
## 🤖 AI Review Summary

**Verdict**: ✅ LGTM

Found **0 errors**, **0 warnings**, **1 suggestion**.

All three files are clean improvements:

- **submit_review.cs**: The null-forgiving operator (`!`) removals are correct per repo convention. Replacing with `?? ""` is safe since `GetString()` on a confirmed `JsonValueKind.String` won’t return null in practice, and the coalescing handles the theoretical case. The `Process.Start` null check is good defensive coding with an actionable error message. Narrowing bare `catch` to `catch (JsonException)` is the right call — `JsonDocument.Parse` only throws `JsonException` for bad input, and the guard `!string.IsNullOrEmpty(stdout)` before the try block means we won’t hit `ArgumentException`. The success-output restructure is cleaner: the user sees confirmation immediately, and the URL is a bonus.

- **review-rules.md**: The new rules are well-scoped and actionable. Thread safety, exception swallowing, pre-allocation, O(n²) warnings, and the testing section are all high-value additions.

- **SKILL.md**: Splitting step 2 (gather context) from step 3 (read PR narrative) is a meaningful workflow improvement — it prevents anchoring bias. The severity levels (❌/⚠️/💡) and verdict line make reviews more scannable.

👍 Clean, focused PR with no scope creep.

- 💡 **Code organization**: Missing postmortem refs line in new Testing section (`review-rules.md:203`)

---

PR #320, opened 2026-04-02
TryGetPlatformDirectoryFromApiLevel: fall back to android-{major}.0

### LINKED ISSUE #319: TryGetPlatformDirectoryFromApiLevel should fall back from android-{major} to android-{major}.0
## Problem

Starting with API 37, Google's SDK Manager installs the platform to `platforms/android-37.0` (package name `platform-37.0_r01`), but `TryGetPlatformDirectoryFromApiLevel` only tries `platforms/android-37` and fails.

This causes **XA5207** for customer projects targeting `net11.0-android37`.

## Current behavior

In [`AndroidSdkInfo.cs`](https://github.com/dotnet/android-tools/blob/ec5040ad5158f240a67d887133dd56cfa5eb74ba/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs#L99-L120):

```csharp
public string? TryGetPlatformDirectoryFromApiLevel (string idOrApiLevel, AndroidVersions versions)
{
    var id = versions.GetIdFromApiLevel (idOrApiLevel);
    string? dir = GetPlatformDirectoryFromId (id);  // \"platforms/android-37\"

    if (Directory.Exists (dir))
        return dir;

    // Existing fallback: minor -> major (e.g. android-36.1 -> android-36)
    // Guarded to NOT fall back for minor versions (dotnet/android#10720)
    if (Version.TryParse (id, out var version) && version.Minor != 0)
        return null;

    // Integer fallback: also tries GetPlatformDirectory(37) -> \"android-37\" (same thing)
    var level = versions.GetApiLevelFromId (id);
    dir = level.HasValue ? GetPlatformDirectory (level.Value) : null;
    if (dir != null && Directory.Exists (dir))
        return dir;

    return null;  // <-- XA5207!
}
```

There is no fallback from `android-{major}` to `android-{major}.0`.

## Proposed fix

When `android-{id}` doesn't exist and `id` is a pure integer (major-only), also try `android-{id}.0`:

```csharp
// After existing checks fail, try major.0 fallback
if (int.TryParse (id, out _)) {
    dir = GetPlatformDirectoryFromId (id + \".0\");
    if (Directory.Exists (dir))
        return dir;
}
```

## Workaround

dotnet/android#11072 aligns the data so that `platformID` is `\"37.0\"` instead of `\"37\"`, which makes the `AndroidApiInfo.xml` `<Id>` element `37.0`. This means the first lookup will try `android-37.0` directly.

However, the android-tools fallback is still valuable as a safety net for cases where the `Id` is just `\"37\"`.

## Context

- dotnet/android#11072 (data fix in dotnet/android)
- dotnet/android#10720 (the PR that added the minor version guard)

jonathanpeppers:
Starting with API 37, Google's SDK Manager installs platform directories as `android-37.0` instead of `android-37`. Both `TryGetPlatformDirectoryFromApiLevel` and `IsPlatformInstalled` only tried `android-{major}`, so projects targeting these newer API levels would fail with **XA5207**.

This adds a `.0` fallback to both methods: when the integer-only directory doesn't exist, also try `android-{id}.0`.

This complements the data-side fix in dotnet/android#11072 (which sets the `Id` to `"37.0"` directly) by providing a safety net when the `Id` is just `"37"`.

### Changes

- **`AndroidSdkInfo.TryGetPlatformDirectoryFromApiLevel`** — After existing lookups fail, try `GetPlatformDirectoryFromId(id + ".0")` when `id` is an integer.
- **`AndroidSdkInfo.IsPlatformInstalled`** — Apply the same `.0` fallback so it stays consistent with `TryGetPlatformDirectoryFromApiLevel`.
- **`AndroidSdkInfoTests`** — New test `TryGetPlatformDirectoryFromApiLevel_MajorFallsBackToMajorDotZero` creates only `android-37.0` and verifies both APIs find it.

Fixes #319

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -120,6 +120,15 @@ namespace Xamarin.Android.Tools
 			if (dir != null && Directory.Exists (dir))
 				return dir;
 
+			// Starting with API 37, Google's SDK Manager installs platforms to
+			// "android-37.0" instead of "android-37". Try the "{major}.0" fallback.
+			// See: https://github.com/dotnet/android-tools/issues/319
+			if (int.TryParse (id, out _)) {
+				dir = GetPlatformDirectoryFromId (id + ".0");
+				if (Directory.Exists (dir))
+					return dir;
+			}
+
 			return null;
 		}
 
--- a/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
@@ -134,7 +134,9 @@ namespace Xamarin.Android.Tools
 
 		public bool IsPlatformInstalled (int apiLevel)
 		{
-			return apiLevel != 0 && Directory.Exists (GetPlatformDirectory (apiLevel));
+			return apiLevel != 0 &&
+				(Directory.Exists (GetPlatformDirectory (apiLevel)) ||
+				 Directory.Exists (GetPlatformDirectoryFromId (apiLevel + ".0")));
 		}
 
 		public string? AndroidNdkPath {
--- a/tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkInfoTests.cs
+++ b/tests/Xamarin.Android.Tools.AndroidSdk-Tests/AndroidSdkInfoTests.cs
@@ -662,6 +662,9 @@ namespace Xamarin.Android.Tools.Tests
 				var dir37 = info.TryGetPlatformDirectoryFromApiLevel ("37", versions);
 				Assert.IsNotNull (dir37, "Should fall back from android-37 to android-37.0");
 				Assert.AreEqual (platform370Path, dir37);
+
+				// IsPlatformInstalled should also find android-37.0
+				Assert.IsTrue (info.IsPlatformInstalled (37), "IsPlatformInstalled should find android-37.0");
 			}
 			finally {
 				Directory.Delete (root, recursive: true);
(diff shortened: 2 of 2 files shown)


src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs
```diff
@@ -120,6 +120,15 @@ public string GetPlatformDirectoryFromId (string id)
 			if (dir != null && Directory.Exists (dir))
 				return dir;
 
+			// Starting with API 37, Google's SDK Manager installs platforms to
+			// "android-37.0" instead of "android-37". Try the "{major}.0" fallback.
+			// See: https://github.com/dotnet/android-tools/issues/319
+			if (int.TryParse (id, out _)) {
```
jonathanpeppers, line 126: 🤖 💡 **Consistency** — `IsPlatformInstalled(int)` also resolves via `GetPlatformDirectory(apiLevel)` → `GetPlatformDirectoryFromId(apiLevel.ToString())`, so `IsPlatformInstalled(37)` would still return `false` when only `android-37.0` is on disk. If any downstream caller gates on `IsPlatformInstalled` before calling `TryGetPlatformDirectoryFromApiLevel`, it would short-circuit and never reach this fallback.

Probably fine given the data-side fix in dotnet/android`#11072` sets `Id = "37.0"` directly, but worth noting for completeness.

_Rule: Consistent behavior across related APIs (Postmortem `#48`)_

jonathanpeppers (commented):
## 🤖 AI Review Summary

**Verdict**: ✅ LGTM

Clean, well-scoped fix for the Google SDK naming change (API 37+). The fallback is correctly placed as a last resort, guarded by `int.TryParse` to prevent triggering on non-integer IDs like `"36.1"` or code names. Good test coverage matching existing patterns.

Found **1 suggestion**:

- 💡 **Consistency**: `IsPlatformInstalled(int)` doesn’t have the `.0` fallback — could return `false` for API 37+ when only `android-37.0` exists (`AndroidSdkInfo.cs`)

👍 Good defensive coding — the `int.TryParse` guard prevents double-dotted fallbacks (`"37.0"` → `"37.0.0"`).

---

PR #327, opened 2026-04-08
Add ADB device tracking (host:track-devices) for real-time device monitoring

### LINKED ISSUE #323: Add ADB device tracking (host:track-devices) for real-time device monitoring
## Summary

Add support for real-time device connection/disconnection monitoring via the ADB daemon socket protocol (`host:track-devices-l`), similar to what the `AndroidSdk.Adbd` NuGet package provides.

## Motivation

The existing `AdbRunner.GetDevicesAsync()` does a one-shot `adb devices -l` call. For IDE extensions and developer tools that need to react immediately when devices connect/disconnect (USB plug, emulator boot, WiFi pairing), polling is suboptimal.

The ADB daemon exposes a `host:track-devices-l` socket command that pushes device list updates in real-time. This is a well-known protocol used by Android Studio and other tooling.

Currently, consumers like MAUI Sherpa depend on the separate `AndroidSdk.Adbd` NuGet (from `redth/android-tools`) for this functionality. Consolidating it into `Xamarin.Android.Tools.AndroidSdk` would reduce the dependency surface.

## Proposed API

`csharp
public class AdbDeviceTracker : IDisposable
{
    public AdbDeviceTracker(string? sdkPath = null, int port = 5037);

    /// Starts tracking. Calls onDevicesChanged whenever the device list changes.
    public Task StartAsync(
        Action<IReadOnlyList<AdbDeviceInfo>> onDevicesChanged,
        CancellationToken cancellationToken = default);

    /// Current snapshot of tracked devices.
    public IReadOnlyList<AdbDeviceInfo> CurrentDevices { get; }

    public void Dispose();
}
`

## Implementation Notes

- Connect to `localhost:5037` (ADB daemon)
- Send `host:track-devices-l` command
- Parse length-prefixed device list updates as they arrive
- Auto-reconnect on connection drops with backoff
- Reuse existing `AdbDeviceInfo` model

## Related

- The `AndroidSdk.Adbd` NuGet (`redth/android-tools`) has a working implementation via `AdbdClient`
- The `maui-labs` CLI does not currently have this feature either

rmarinho:
## Summary

Add support for real-time device connection/disconnection monitoring via the ADB daemon socket protocol (`host:track-devices-l`).

### Changes
- Added `AdbDeviceTracker` class implementing `IDisposable`
- Added `AdbClient` internal class encapsulating ADB daemon TCP socket protocol
- Socket connection to `localhost:5037` (ADB daemon)
- Sends `host:track-devices-l` command and reads length-prefixed device list updates
- Callback-based `StartAsync()` with `CancellationToken` support
- `CurrentDevices` snapshot property for current device state
- Auto-reconnect with exponential backoff (500ms to 16s) on connection drops
- Reuses existing `AdbRunner.ParseAdbDevicesOutput()` for parsing
- Unit tests covering protocol parsing, lifecycle, and edge cases
- Updated `PublicAPI.Unshipped.txt` for both `net10.0` and `netstandard2.0`

### API
```csharp
public sealed class AdbDeviceTracker : IDisposable
{
    public AdbDeviceTracker(int port = 5037, Action<TraceLevel, string>? logger = null);
    public IReadOnlyList<AdbDeviceInfo> CurrentDevices { get; }
    public Task StartAsync(Action<IReadOnlyList<AdbDeviceInfo>> onDevicesChanged, CancellationToken cancellationToken = default);
    public void Dispose();
}
```

Closes #323

### DIFF
--- a/src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
@@ -14,18 +14,14 @@ namespace Xamarin.Android.Tools;
 /// Low-level ADB daemon socket protocol client.
 /// Encapsulates a single TCP connection to the ADB server and exposes
 /// the wire protocol operations (send command, read status, read length-prefixed payloads).
-/// One instance = one connection. Dispose closes the socket.
+/// One instance can be reused across reconnections via <see cref="ReconnectAsync"/>.
+/// Dispose closes the socket.
 /// </summary>
 internal sealed class AdbClient : IDisposable
 {
-	readonly TcpClient client;
+	TcpClient? client;
 	NetworkStream? stream;
 	bool disposed;
-
-	public AdbClient ()
-	{
-		client = new TcpClient ();
-	}
 
 	/// <summary>
 	/// Connects to the ADB daemon at 127.0.0.1 on the specified port.
@@ -33,13 +29,24 @@ internal sealed class AdbClient : IDisposable
 	public async Task ConnectAsync (int port, CancellationToken cancellationToken = default)
 	{
 		ThrowIfDisposed ();
+		var tcp = new TcpClient ();
+		client = tcp;
 #if NET5_0_OR_GREATER
-		await client.ConnectAsync ("127.0.0.1", port, cancellationToken).ConfigureAwait (false);
+		await tcp.ConnectAsync ("127.0.0.1", port, cancellationToken).ConfigureAwait (false);
 #else
-		await client.ConnectAsync ("127.0.0.1", port).ConfigureAwait (false);
+		await tcp.ConnectAsync ("127.0.0.1", port).ConfigureAwait (false);
 		cancellationToken.ThrowIfCancellationRequested ();
 #endif
-		stream = client.GetStream ();
+		stream = tcp.GetStream ();
+	}
+
+	/// <summary>
+	/// Closes the current connection and establishes a new one.
+	/// </summary>
+	public async Task ReconnectAsync (int port, CancellationToken cancellationToken = default)
+	{
+		CloseConnection ();
+		await ConnectAsync (port, cancellationToken).ConfigureAwait (false);
 	}
 
 	/// <summary>
@@ -49,29 +56,37 @@ internal sealed class AdbClient : IDisposable
 	public async Task SendCommandAsync (string command, CancellationToken cancellationToken = default)
 	{
 		var s = GetStream ();
+		// Combine length prefix + command into a single write to minimize syscalls
 		var commandBytes = Encoding.ASCII.GetBytes (command);
-		var header = commandBytes.Length.ToString ("x4");
-		var headerBytes = Encoding.ASCII.GetBytes (header);
-
-		await s.WriteAsync (headerBytes, 0, headerBytes.Length, cancellationToken).ConfigureAwait (false);
-		await s.WriteAsync (commandBytes, 0, commandBytes.Length, cancellationToken).ConfigureAwait (false);
+		var lengthPrefix = commandBytes.Length.ToString ("x4");
+		var packet = new byte [4 + commandBytes.Length];
+		Encoding.ASCII.GetBytes (lengthPrefix, 0, 4, packet, 0);
+		Buffer.BlockCopy (commandBytes, 0, packet, 4, commandBytes.Length);
+#if NET5_0_OR_GREATER
+		await s.WriteAsync (packet.AsMemory (), cancellationToken).ConfigureAwait (false);
+#else
+		await s.WriteAsync (packet, 0, packet.Length, cancellationToken).ConfigureAwait (false);
+#endif
 		await s.FlushAsync (cancellationToken).ConfigureAwait (false);
 	}
 
 	/// <summary>
 	/// Reads the 4-byte status response from the ADB daemon.
-	/// Returns <see cref="AdbResponseStatus.Okay"/> or <see cref="AdbResponseStatus.Fail"/>.
 	/// </summary>
 	public async Task<AdbResponseStatus> ReadStatusAsync (CancellationToken cancellationToken = default)
 	{
 		var s = GetStream ();
-		var statusBytes = await ReadExactBytesAsync (s, 4, cancellationToken).ConfigureAwait (false);
+		var statusBytes = await ReadExactBytesIntoNewArrayAsync (s, 4, cancellationToken).ConfigureAwait (false);
+		// Status is always 4 ASCII chars
+		if (statusBytes [0] == (byte) 'O' && statusBytes [1] == (byte) 'K' &&
+			statusBytes [2] == (byte) 'A' && statusBytes [3] == (byte) 'Y')
+			return AdbResponseStatus.Okay;
+		if (statusBytes [0] == (byte) 'F' && statusBytes [1] == (byte) 'A' &&
+			statusBytes [2] == (byte) 'I' && statusBytes [3] == (byte) 'L')
+			return AdbResponseStatus.Fail;
+
 		var status = Encoding.ASCII.GetString (statusBytes, 0, 4);
-		return status switch {
-			"OKAY" => AdbResponseStatus.Okay,
-			"FAIL" => AdbResponseStatus.Fail,
-			_ => throw new InvalidOperationException ($"Unexpected ADB status: '{status}'"),
-		};
+		throw new InvalidOperationException ($"Unexpected ADB status: '{status}'");
 	}
 
 	/// <summary>
@@ -100,10 +115,8 @@ internal sealed class AdbClient : IDisposable
 	/// </summary>
 	public async Task<string?> ReadLengthPrefixedStringAsync (CancellationToken cancellationToken = default)
 	{
-		var bytes = await ReadLengthPrefixedBytesAsync (cancellationToken).ConfigureAwait (false);
-		if (bytes == null)
-			return null;
-		return Encoding.ASCII.GetString (bytes, 0, bytes.Length);
+		var s = GetStream ();
+		return await ReadLengthPrefixedStringCoreAsync (s, cancellationToken).ConfigureAwait (false);
 	}
 
 	/// <summary>
@@ -113,18 +126,7 @@ internal sealed class AdbClient : IDisposable
 	public async Task<byte[]?> ReadLengthPrefixedBytesAsync (CancellationToken cancellationToken = default)
 	{
 		var s = GetStream ();
-		var lengthBytes = await ReadExactBytesOrNullAsync (s, 4, cancellationToken).ConfigureAwait (false);
-		if (lengthBytes == null)
-			return null;
-
-		var lengthHex = Encoding.ASCII.GetString (lengthBytes, 0, 4);
-		if (!int.TryParse (lengthHex, System.Globalization.NumberStyles.HexNumber, null, out var length))
-			throw new FormatException ($"Invalid ADB length prefix: '{lengthHex}'");
-
-		if (length == 0)
-			return Array.Empty<byte> ();
-
-		return await ReadExactBytesAsync (s, length, cancellationToken).ConfigureAwait (false);
+		return await ReadLengthPrefixedBytesCoreAsync (s, cancellationToken).ConfigureAwait (false);
 	}
 
 	/// <summary>
@@ -132,7 +134,7 @@ internal sealed class AdbClient : IDisposable
 	/// </summary>
 	public void Close ()
 	{
-		client.Close ();
+		CloseConnection ();
 	}
 
 	public void Dispose ()
@@ -140,8 +142,18 @@ internal sealed class AdbClient : IDisposable
 		if (disposed)
 			return;
 		disposed = true;
-		client.Close ();
-		client.Dispose ();
+		CloseConnection ();
+	}
+
+	void CloseConnection ()
+	{
+		stream = null;
+		var tcp = client;
+		client = null;
+		if (tcp != null) {
+			tcp.Close ();
+			tcp.Dispose ();
+		}
 	}
 
 	NetworkStream GetStream ()
@@ -158,11 +170,26 @@ internal sealed class AdbClient : IDisposable
 			throw new ObjectDisposedException (nameof (AdbClient));
 	}
 
+	// --- Shared core implementations (used by both instance and static methods) ---
+
 	/// <summary>
 	/// Reads a length-prefixed ASCII string from a raw stream.
-	/// Useful for testing and for callers that already have a stream.
+	/// Shared core for both instance and static access patterns.
 	/// </summary>
 	internal static async Task<string?> ReadLengthPrefixedStringFromStreamAsync (Stream stream, CancellationToken cancellationToken)
+	{
+		return await ReadLengthPrefixedStringCoreAsync (stream, cancellationToken).ConfigureAwait (false);
+	}
+
+	static async Task<string?> ReadLengthPrefixedStringCoreAsync (Stream stream, CancellationToken cancellationToken)
+	{
+		var bytes = await ReadLengthPrefixedBytesCoreAsync (stream, cancellationToken).ConfigureAwait (false);
+		if (bytes == null)
+			return null;
+		return Encoding.ASCII.GetString (bytes, 0, bytes.Length);
+	}
+
+	static async Task<byte[]?> ReadLengthPrefixedBytesCoreAsync (Stream stream, CancellationToken cancellationToken)
 	{
 		var lengthBytes = await ReadExactBytesOrNullAsync (stream, 4, cancellationToken).ConfigureAwait (false);
 		if (lengthBytes == null)
@@ -173,13 +200,12 @@ internal sealed class AdbClient : IDisposable
 			throw new FormatException ($"Invalid ADB length prefix: '{lengthHex}'");
 
 		if (length == 0)
-			return string.Empty;
+			return Array.Empty<byte> ();
 
-		var payload = await ReadExactBytesAsync (stream, length, cancellationToken).ConfigureAwait (false);
-		return Encoding.ASCII.GetString (payload, 0, payload.Length);
+		return await ReadExactBytesIntoNewArrayAsync (stream, length, cancellationToken).ConfigureAwait (false);
 	}
 
-	static async Task<byte[]> ReadExactBytesAsync (Stream stream, int count, CancellationToken cancellationToken)
+	static async Task<byte[]> ReadExactBytesIntoNewArrayAsync (Stream stream, int count, CancellationToken cancellationToken)
 	{
 		var result = await ReadExactBytesOrNullAsync (stream, count, cancellationToken).ConfigureAwait (false);
 		if (result == null)
--- a/src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
+++ b/src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Buffers;
 using System.IO;
 using System.Net.Sockets;
 using System.Text;
@@ -17,8 +18,14 @@ namespace Xamarin.Android.Tools;
 /// One instance can be reused across reconnections via <see cref="ReconnectAsync"/>.
 /// Dispose closes the socket.
 /// </summary>
+/// <remarks>
+/// This class is not thread-safe. All protocol operations must be serialized by the caller.
+/// </remarks>
 internal sealed class AdbClient : IDisposable
 {
+	// Reusable 4-byte buffer for status/length reads (safe: single-caller, non-concurrent)
+	readonly byte[] headerBuffer = new byte [4];
+
 	TcpClient? client;
 	NetworkStream? stream;
 	bool disposed;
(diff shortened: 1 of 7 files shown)


src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
```diff
@@ -0,0 +1,188 @@
...
+		Action<IReadOnlyList<AdbDeviceInfo>> onDevicesChanged,
+		CancellationToken cancellationToken)
+	{
+		using var client = new TcpClient ();
+#if NET5_0_OR_GREATER
+		await client.ConnectAsync ("127.0.0.1", port, cancellationToken).ConfigureAwait (false);
+#else
+		await client.ConnectAsync ("127.0.0.1", port).ConfigureAwait (false);
+		cancellationToken.ThrowIfCancellationRequested ();
+#endif
```
jonathanpeppers: If we are going to add a `TcpClient` to work with `adb`'s daemon directly.

I'd prefer us to use an `internal AdbClient` class we can start sharing with other code in this library.

If we have this, we can avoid shelling out to `adb.exe` and starting a new process for every single operation.

You might start using this with some of the existing code, too. Is there a place we run `adb.exe` a lot in the recent changes?

rmarinho: Done! Extracted an `internal AdbClient` class that encapsulates the ADB daemon TCP socket protocol — connect, send length-prefixed commands, read OKAY/FAIL status, and read length-prefixed payloads. It's byte-oriented at the transport layer with string helpers on top for ASCII use cases.

`AdbDeviceTracker` now uses `AdbClient` via composition, which makes `TrackDevicesAsync` much cleaner and intent-clear.

Re: reuse — yes, `AdbRunner` shells out to `adb.exe` for every operation (`ListDevicesAsync`, `GetShellPropertyAsync`, `ReversePortAsync`, `ListReversePortsAsync`, etc.). We can incrementally migrate those to use `AdbClient` where the daemon protocol supports it (e.g. `host:devices-l`, `host:forward`, `host:kill`). I'll look at which ones are hot paths and prioritize those in a follow-up.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+	public async Task SendCommandAsync (string command, CancellationToken cancellationToken = default)
+	{
+		var s = GetStream ();
+		var commandBytes = Encoding.ASCII.GetBytes (command);
+		var header = commandBytes.Length.ToString ("x4");
+		var headerBytes = Encoding.ASCII.GetBytes (header);
+
+		await s.WriteAsync (headerBytes, 0, headerBytes.Length, cancellationToken).ConfigureAwait (false);
+		await s.WriteAsync (commandBytes, 0, commandBytes.Length, cancellationToken).ConfigureAwait (false);
+		await s.FlushAsync (cancellationToken).ConfigureAwait (false);
```
jonathanpeppers: There is an analyzer that requires `ConfigureAwait(false)`, if we need that everywhere, should we enable it?
* https://github.com/dotnet/sdk/pull/54116

We can do this in a separate PR.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+
+	/// <summary>
+	/// Reads the 4-byte status response from the ADB daemon.
+	/// Returns <see cref="AdbResponseStatus.Okay"/> or <see cref="AdbResponseStatus.Fail"/>.
+	/// </summary>
+	public async Task<AdbResponseStatus> ReadStatusAsync (CancellationToken cancellationToken = default)
+	{
+		var s = GetStream ();
+		var statusBytes = await ReadExactBytesAsync (s, 4, cancellationToken).ConfigureAwait (false);
+		var status = Encoding.ASCII.GetString (statusBytes, 0, 4);
```
jonathanpeppers: Is there an overload where you can pass in a `Span<byte> buffer = stackalloc byte[4];`

This would prevent every call allocating a `byte[]` on the heap.

There are several examples in this file:

https://github.com/dotnet/android-tools/blob/ed6aab1e3675ac2f9e1e928c424c78de1faff674/src/Microsoft.Android.Build.BaseTasks/Files.cs#L540-L542

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+	/// Returns null if the connection is closed cleanly before the length prefix.
+	/// </summary>
+	public async Task<byte[]?> ReadLengthPrefixedBytesAsync (CancellationToken cancellationToken = default)
+	{
+		var s = GetStream ();
+		var lengthBytes = await ReadExactBytesOrNullAsync (s, 4, cancellationToken).ConfigureAwait (false);
+		if (lengthBytes == null)
+			return null;
+
+		var lengthHex = Encoding.ASCII.GetString (lengthBytes, 0, 4);
```
jonathanpeppers: Same here, we allocate a `byte[]`, where we could use `stackalloc` + Span.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+	/// </summary>
+	internal static async Task<string?> ReadLengthPrefixedStringFromStreamAsync (Stream stream, CancellationToken cancellationToken)
+	{
+		var lengthBytes = await ReadExactBytesOrNullAsync (stream, 4, cancellationToken).ConfigureAwait (false);
+		if (lengthBytes == null)
+			return null;
+
+		var lengthHex = Encoding.ASCII.GetString (lengthBytes, 0, 4);
+		if (!int.TryParse (lengthHex, System.Globalization.NumberStyles.HexNumber, null, out var length))
+			throw new FormatException ($"Invalid ADB length prefix: '{lengthHex}'");
```
jonathanpeppers: See if the AI will go through the whole file allocating as few `byte[]` as possible, using the techniques in `Files.cs`.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+	{
+		var result = await ReadExactBytesOrNullAsync (stream, count, cancellationToken).ConfigureAwait (false);
+		if (result == null)
+			throw new IOException ($"Unexpected end of stream (expected {count} bytes).");
+		return result;
+	}
+
+	static async Task<byte[]?> ReadExactBytesOrNullAsync (Stream stream, int count, CancellationToken cancellationToken)
+	{
+		var buffer = new byte [count];
```
jonathanpeppers: Here is another, it could check the count against a "max size" and use `stackalloc` for small sizes.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
```diff
@@ -0,0 +1,154 @@
...
+/// Monitors ADB device connections in real-time via the <c>host:track-devices-l</c> socket protocol.
+/// Pushes device list updates through a callback whenever devices connect, disconnect, or change state.
+/// </summary>
+public sealed class AdbDeviceTracker : IDisposable
+{
+	readonly object syncLock = new object ();
+	readonly int port;
+	readonly Action<TraceLevel, string> logger;
+	volatile IReadOnlyList<AdbDeviceInfo> currentDevices = Array.Empty<AdbDeviceInfo> ();
+	AdbClient? activeClient;
```
jonathanpeppers: Why does this class need to support more than one `AdbClient`? should it just reuse a single one for its lifetime?

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: Created #353 to track this.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
rmarinho: Fixed — `AdbDeviceTracker` now creates a single `AdbClient` at construction and reuses it via `ReconnectAsync()` on each reconnection attempt. No more per-iteration `new AdbClient()`.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: Good point. Unfortunately `stackalloc` can't be used across `await` boundaries (the compiler prohibits it in async methods), and `Stream.ReadAsync` requires either `byte[]` or `Memory<byte>` — neither works with stack-allocated memory.

What I did instead:
- `SendCommandAsync`: combines the 4-byte prefix + command into a single `byte[]` and single `WriteAsync` call (one allocation, one syscall instead of two)
- `ReadStatusAsync`: compares raw bytes directly (`'O','K','A','Y'`) — avoids the `Encoding.ASCII.GetString` allocation for the common path
- The 4-byte reads in `ReadExactBytesOrNullAsync` still need a heap `byte[4]` because of the async constraint, but it's minimal

For truly zero-alloc small reads, we'd need a sync code path or `PipeReader` (which doesn't multi-target well to netstandard2.0). Happy to explore that as a perf follow-up if this becomes hot.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: Ok did another pass with that

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
```diff
@@ -0,0 +1,189 @@
...
+	public AdbDeviceTracker (string? adbPath = null, int port = 5037,
+		IDictionary<string, string>? environmentVariables = null,
+		Action<TraceLevel, string>? logger = null)
+	{
+		if (port <= 0 || port > 65535)
+			throw new ArgumentOutOfRangeException (nameof (port), "Port must be between 1 and 65535.");
+		this.adbPath = adbPath;
+		this.port = port;
+		this.environmentVariables = environmentVariables;
+		this.logger = logger ?? RunnerDefaults.NullLogger;
```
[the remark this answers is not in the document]
rmarinho: The unused `adbPath` / `environmentVariables` constructor parameters were already removed in commit 544bee5 — the current ctor is `AdbDeviceTracker(int port = 5037, Action<TraceLevel, string>? logger = null)`, so CS0414 no longer fires. The tracker uses pure TCP wire protocol against the ADB daemon and doesn't need a process path; callers that want to ensure the daemon is running can use `AdbRunner` (which does take `adbPath` + `environmentVariables`) before constructing a tracker. Leaving the smaller surface in place. Pushing 7d5d7a2 also addresses the other two threads.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,210 @@
...
+		var lengthHex = Encoding.ASCII.GetString (lengthBytes, 0, 4);
+		if (!int.TryParse (lengthHex, System.Globalization.NumberStyles.HexNumber, null, out var length))
+			throw new FormatException ($"Invalid ADB length prefix: '{lengthHex}'");
+
+		if (length == 0)
+			return string.Empty;
+
+		var payload = await ReadExactBytesAsync (stream, length, cancellationToken).ConfigureAwait (false);
+		return Encoding.ASCII.GetString (payload, 0, payload.Length);
+	}
```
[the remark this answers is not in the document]
rmarinho: Fixed in 7d5d7a2 by extracting a private shared core:

```csharp
static async Task<byte[]?> ReadLengthPrefixedBytesFromStreamAsync (
    Stream stream, byte[] lengthBuffer, CancellationToken cancellationToken)
```

- The instance `ReadLengthPrefixedBytesAsync` calls the core with the reusable `headerBuffer` field (pooled-buffer optimization preserved).
- The static `ReadLengthPrefixedStringFromStreamAsync` (test API, signature unchanged) allocates a fresh 4-byte buffer, calls the core, and ASCII-decodes.
- The instance `ReadLengthPrefixedStringAsync` now delegates to the instance bytes method and decodes, so the ASCII conversion lives in one place.

All `AdbClientTests` (including null-on-EOF, empty-payload, truncated-payload, and invalid-hex) still pass without changes.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
```diff
@@ -0,0 +1,154 @@
...
+				} catch (Exception ex) when (ex is IOException || ex is SocketException || ex is ObjectDisposedException) {
+					if (token.IsCancellationRequested)
+						break;
+					logger.Invoke (TraceLevel.Warning, $"ADB tracking connection lost: {ex.Message}. Reconnecting in {backoffMs}ms...");
+					try {
+						await Task.Delay (backoffMs, token).ConfigureAwait (false);
+					} catch (OperationCanceledException) {
+						break;
+					}
+					backoffMs = Math.Min (backoffMs * 2, MaxBackoffMs);
```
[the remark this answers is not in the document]
rmarinho: Real bug — fixed in 7d5d7a2. You were right that the previous `backoffMs = InitialBackoffMs;` after the inner try/catch was effectively dead code, because the read loop in `TrackDevicesAsync` only exits via exception or cancellation.

`TrackDevicesAsync` is now split into two methods:

- `ConnectAndHandshakeAsync` — `ReconnectAsync` + `SendCommandAsync("host:track-devices-l")` + `EnsureOkayAsync`.
- `ReadTrackingUpdatesAsync` — the infinite length-prefixed read loop.

`StartAsync` resets `backoffMs = InitialBackoffMs` *between* the two calls, so the reset only happens after a fully successful protocol handshake (not after a bare TCP connect, and not before an attempt that might fail mid-handshake). A long-lived session that drops after hours will now retry fresh from 500ms instead of using the accumulated max backoff.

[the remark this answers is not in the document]
rmarinho: Follow-up in bed643b based on independent reviews by Claude Opus 4.8 and GPT-5.5 — both flagged that 7d5d7a2''s reset boundary was still too eager.

Resetting `backoffMs` after `EnsureOkayAsync` means a daemon that accepts the TCP connection and answers `OKAY`, then drops the socket before the first device-list payload arrives, would pin reconnects at `InitialBackoffMs` (500 ms) forever instead of climbing toward `MaxBackoffMs`. The reset masked exactly the flap that the exponential backoff was meant to throttle.

Fixed by moving the reset into `ReadTrackingUpdatesAsync` via an `onConnectionStable` callback that fires after the **first successful payload read** — so the reset condition is "we proved this session is actually usable" rather than "the handshake said OKAY". `host:track-devices-l` always pushes the full device list immediately on connect, so under healthy conditions the reset still fires within milliseconds.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,303 @@
...
+/// One instance can be reused across reconnections via <see cref="ReconnectAsync"/>.
+/// Dispose closes the socket.
+/// </summary>
+/// <remarks>
+/// This class is not thread-safe. All protocol operations must be serialized by the caller.
+/// </remarks>
+internal sealed class AdbClient : IDisposable
+{
+	// Reusable 4-byte buffer for status/length reads (safe: single-caller, non-concurrent)
+	readonly byte[] headerBuffer = new byte [4];
```
jonathanpeppers: Why does this think it's OK and "non-concurrent"? LOL?

It looks highly concurrent!

Can you just use `stackalloc` instead? That should be fine for 4 bytes.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	void ThrowIfDisposed ()
+	{
+		if (disposed)
+			throw new ObjectDisposedException (nameof (AdbClient));
+	}
+
+	// --- Shared core implementations (used by static method for tests) ---
+
```
jonathanpeppers: `#region` in disguise!

```suggestion
```

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,303 @@
...
+					return false;
+				throw new IOException ($"Unexpected end of stream (read {totalRead} of {count} bytes).");
+			}
+			totalRead += read;
+		}
+		return true;
+	}
+
+	// --- Hex encoding/decoding helpers (avoid string allocations) ---
+
```
jonathanpeppers:
```suggestion
```

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,303 @@
...
+	/// <summary>
+	/// Writes a 4-digit lowercase hex representation of <paramref name="value"/> into the first 4 bytes of <paramref name="buffer"/>.
+	/// </summary>
+	static void WriteHexLength (byte[] buffer, int value)
+	{
+		buffer [0] = HexChars [(value >> 12) & 0xF];
+		buffer [1] = HexChars [(value >> 8) & 0xF];
+		buffer [2] = HexChars [(value >> 4) & 0xF];
+		buffer [3] = HexChars [value & 0xF];
+	}
```
jonathanpeppers: Can we add an `internal` overload of these:
https://github.com/dotnet/android-tools/blob/06a086724d3c65bf58b0cf248940d4f5115aa95b/src/Microsoft.Android.Build.BaseTasks/Files.cs#L588-L608

Where you pass in a `byte[]` and try to reuse as much existing code as possible?

I am not a fan of that `HexChars` array above.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
```diff
@@ -0,0 +1,303 @@
...
+			else if (b >= (byte) 'a' && b <= (byte) 'f')
+				nibble = b - 'a' + 10;
+			else if (b >= (byte) 'A' && b <= (byte) 'F')
+				nibble = b - 'A' + 10;
+			else
+				throw new FormatException ($"Invalid ADB length prefix: '{Encoding.ASCII.GetString (buffer, 0, 4)}'");
+			value = (value << 4) | nibble;
+		}
+		return value;
+	}
```
jonathanpeppers, line 294: Same with this one, should maybe be a new `internal` method in the `Files` class and reuse existing code.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbDeviceTracker.cs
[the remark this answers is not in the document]
rmarinho: The unused `adbPath` / `environmentVariables` surface was removed in the current branch, so this constructor now matches the actual TCP-only tracker behavior. Callers that need to ensure the daemon is running can do that separately via `AdbRunner`.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: test

rmarinho: test reply

rmarinho: The shared 4-byte buffer is gone now. The async read path uses a local byte[4] buffer per read, which avoids the shared mutable-state concern and still works across await. stackalloc is not viable here because Stream.ReadAsync needs a real buffer across an async boundary.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: The section markers are gone as well; the read helpers now just contain the minimal implementation the file needs.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: I replaced the lookup-table approach with a tiny nibble helper, so the hex write path no longer depends on a separate HexChars array.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: The backoff reset now happens immediately after a successful connection/handshake, and the old reset path in the retry loop was removed because that path was never reached in normal operation.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbClient.cs
rmarinho: I verified the current tracker path and kept the constructor intentionally minimal, which avoids the unused-field warning without adding process-launch plumbing the tracker does not use.

PR #328, opened 2026-04-13
Validate extracted zip entry paths in Files.ExtractAll

jonathanpeppers:
Ensure resolved output paths stay within the destination directory before extracting zip entries. This is just for correctness; there's no security enforcement here. Entries with path traversal (e.g. via `modifyCallback` introducing `../`) are skipped with a warning logged via the new `TaskLoggingHelper` parameter.

Note that libZipSharp already normalizes entry names on read:

```csharp
var stream = new MemoryStream();
using (var sysZip = new System.IO.Compression.ZipArchive(stream,
    System.IO.Compression.ZipArchiveMode.Create, leaveOpen: true))
{
    sysZip.CreateEntry("../relative.txt");
}
stream.Position = 0;
using (var zip = Xamarin.Tools.Zip.ZipArchive.Open(stream))
{
    foreach (var entry in zip)
        Console.WriteLine(entry.FullName); // prints 'evil.txt'
}
```

The original overload without `TaskLoggingHelper` is preserved and marked `[Obsolete]` to maintain binary compatibility.

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/Files.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/Files.cs
@@ -441,14 +441,22 @@ namespace Microsoft.Android.Build.Tasks
 			}
 		}
 
+		[Obsolete ("Use the overload that accepts a TaskLoggingHelper parameter.")]
+		public static bool ExtractAll (ZipArchive zip, string destination, Action<int, int> progressCallback,
+			Func<string, string> modifyCallback, Func<string, bool> deleteCallback, Func<string, bool> skipCallback)
+		{
+			return ExtractAll (zip, destination, progressCallback, modifyCallback, deleteCallback, skipCallback, log: null);
+		}
+
 		public static bool ExtractAll (ZipArchive zip, string destination, Action<int, int> progressCallback = null, Func<string, string> modifyCallback = null,
-			Func<string, bool> deleteCallback = null, Func<string, bool> skipCallback = null)
+			Func<string, bool> deleteCallback = null, Func<string, bool> skipCallback = null, TaskLoggingHelper log = null)
 		{
 			int i = 0;
 			int total = (int)zip.EntryCount;
 			bool updated = false;
 			var files = new HashSet<string> ();
 			var memoryStream = MemoryStreamPool.Shared.Rent ();
+			var fullDestination = Path.GetFullPath (destination + Path.DirectorySeparatorChar);
 			try {
 				foreach (var entry in zip) {
 					progressCallback?.Invoke (i++, total);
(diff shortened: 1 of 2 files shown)


jonathanpeppers (commented):
## 🤖 AI Review Summary

**Verdict**: ✅ LGTM

Good defense-in-depth for zip-slip via `modifyCallback`. The `Path.GetFullPath` + trailing directory separator + `StartsWith` approach is correct and avoids both path traversal and the `C:\ProgramFilesX` false-match pitfall. Binary compat with the `[Obsolete]` overload is sound.

Found **1 issue**:

- ⚠️ **Performance**: `fullDestination` is loop-invariant but recomputed every iteration (`Files.cs:473`)
- 💡 **Testing**: consider additional edge-case coverage (`FilesTests.cs:629`)

👍 Honest PR description acknowledging libZipSharp already normalizes entry names, correctly scoping this as a guard for `modifyCallback`-introduced traversal.

---

src/Microsoft.Android.Build.BaseTasks/Files.cs
```diff
@@ -463,6 +470,11 @@ public static bool ExtractAll (ZipArchive zip, string destination, Action<int, i
 						continue;
 					var fullName = modifyCallback?.Invoke (entry.FullName) ?? entry.FullName;
 					var outfile = Path.GetFullPath (Path.Combine (destination, fullName));
+					var fullDestination = Path.GetFullPath (destination + Path.DirectorySeparatorChar);
```
jonathanpeppers: 🤖 ⚠️ **Performance** — `fullDestination` depends only on `destination`, which never changes across iterations. Hoist it above the `foreach` loop to avoid repeated `string` concatenation + `Path.GetFullPath()` on every zip entry.

```csharp
var fullDestination = Path.GetFullPath (destination + Path.DirectorySeparatorChar);
```

Move this to just before `foreach (var entry in zip)` (around the `try {` block).

_Rule: Place cheap checks before expensive ones; avoid repeated work in loops (Postmortem `#21`)_

tests/Microsoft.Android.Build.BaseTasks-Tests/FilesTests.cs
```diff
@@ -625,6 +625,23 @@ public void ExtractAll_MacOSFiles ()
 			DirectoryAssert.DoesNotExist (tempDir);
 		}
 
+		[Test]
+		public void ExtractAll_ZipSlip_SkipsPathTraversal ()
```
jonathanpeppers: 🤖 💡 **Testing** — Consider adding a test case where only *some* entries trigger traversal, verifying that valid entries are still extracted while traversal entries are skipped. This catches regressions where the `continue` inadvertently skips all subsequent entries.

Also worth a `[TestCase]` with `..\\` (Windows-style separators) and deeply nested `../../..` to cover cross-platform edge cases.

_Rule: Test edge cases — if the PR only tests the happy path, suggest edge cases (Postmortem `#13`)_

simonrozsival (commented):
This fix looks reasonable

PR #335, opened 2026-04-22
Replace Crc64 with XxHash64 in Files class

jonathanpeppers:
Replaces the software `Crc64` implementation with hardware-accelerated `XxHash64` from `System.IO.Hashing` 10.0.7 for all hashing in the `Files` class (`HashBytes`, `HashFile`, `HashStream`).

The `Crc64` class is retained — only its usage in `Files` is replaced.

## Changes

- **`Files.cs`**: `HashBytes`, `HashFile(string)`, and `HashStream` now use `XxHash64` instead of `Crc64`. Uses `stackalloc` with Span-based APIs for zero-allocation hashing.  `HashFile(string, HashAlgorithm)` overload is unchanged for backward compat.
- **`Files.cs`**: `ToHexString(ReadOnlySpan<byte>)` is now `public`. Uses `stackalloc` for small hashes (≤64 bytes) to avoid `char[]` allocation. The `byte[]` overload delegates to it.
- **`Files.cs`**: `HashFile` overloads simplified to use `File.OpenRead()`.
- **`Microsoft.Android.Build.BaseTasks.csproj`**: Added `System.IO.Hashing` package reference.
- **`Directory.Build.targets`**: Pinned `System.IO.Hashing` version to 10.0.7 (centralized, matching other packages).
- **`FilesTests.cs`**: Fixed pre-existing bug where tests reassigned `tempDir` from a directory to a file path, causing `TearDown` to skip cleanup. All tests now use `tempDir` as a directory and write to local `tempFile` variables inside it.

## Benchmark results (1 MB data, Intel i9-14900KF)

| Method         | Before (Crc64) | Alloc | After (XxHash64) | Alloc | Speedup |
|--------------- |---------------:|------:|------------------:|------:|--------:|
| HashBytes      |       468.2 μs | 232 B |          46.00 μs |  56 B |   ~10x  |
| HashStream     |       482.9 μs | 232 B |          44.98 μs | 184 B |   ~11x  |
| HashFile       |       764.9 μs | 474 B |          75.83 μs | 424 B |   ~10x  |
| HasFileChanged |     1,496.6 μs | 945 B |         163.23 μs | 848 B |    ~9x  |

Depends on #334

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -28,6 +28,7 @@
     <PackageReference Update="nunit"                                        Version="3.13.2" />
     <PackageReference Update="NUnit3TestAdapter"                            Version="4.0.0" />
     <PackageReference Update="BenchmarkDotNet"                              Version="0.14.0" />
+    <PackageReference Update="System.IO.Hashing"                            Version="10.0.7" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Android.Build.BaseTasks/Files.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/Files.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Hashing;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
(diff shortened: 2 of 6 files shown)


simonrozsival:
That's very interesting build perf improvement. I will switch to xxhash64 in the trimmable type map.

simonrozsival (approved):

PR #342, opened 2026-04-24
Fix NullReferenceException in DeleteFile catch block

jonathanpeppers:
## Summary

In `Files.DeleteFile()`, the catch block used `var helper = log as TaskLoggingHelper` followed by an unconditional dereference of `helper`. Since the `log` parameter is typed `object`, when it is `null` or not a `TaskLoggingHelper`, the `as` cast returns `null` and the next line throws a `NullReferenceException` — masking the original file deletion error with a confusing unrelated exception.

## Changes

- **`src/Microsoft.Android.Build.BaseTasks/Files.cs`** — replaced the unsafe `as` cast + dereference with pattern matching: `if (log is TaskLoggingHelper helper) helper.LogErrorFromException(ex);`. When `log` is null or the wrong type, the deletion exception is now silently caught (preserving the existing API contract that `DeleteFile` does not throw).

- **`tests/Microsoft.Android.Build.BaseTasks-Tests/FilesTests.cs`** — added two new tests:
  - `DeleteFile_NullLog_DoesNotThrow` — verifies no exception when `log` is `null`
  - `DeleteFile_NonTaskLoggingHelperLog_DoesNotThrow` — verifies no exception when `log` is a non-`TaskLoggingHelper` object

## Verification

- ✅ `dotnet build Xamarin.Android.Tools.sln` — clean (0 errors)
- ✅ `dotnet test tests/Microsoft.Android.Build.BaseTasks-Tests` — all 103 tests pass

### DIFF
--- a/src/Microsoft.Android.Build.BaseTasks/Files.cs
+++ b/src/Microsoft.Android.Build.BaseTasks/Files.cs
@@ -597,8 +597,9 @@ namespace Microsoft.Android.Build.Tasks
 			try {
 				File.Delete (filename);
 			} catch (Exception ex) {
-				var helper = log as TaskLoggingHelper;
-				helper.LogErrorFromException (ex);
+				if (log is TaskLoggingHelper helper) {
+					helper.LogErrorFromException (ex);
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Android.Build.BaseTasks/Files.cs
```diff
@@ -597,8 +597,9 @@ public static void DeleteFile (string filename, object log)
 			try {
 				File.Delete (filename);
 			} catch (Exception ex) {
-				var helper = log as TaskLoggingHelper;
-				helper.LogErrorFromException (ex);
+				if (log is TaskLoggingHelper helper) {
+					helper.LogErrorFromException (ex);
+				}
```
jonathanpeppers, line 602: Since AI found this, filed to opt into NRT project-wide:
* https://github.com/dotnet/android-tools/issues/343

simonrozsival (approved):

PR #305, opened 2026-03-13
Add ADB reverse port forwarding support

rmarinho:
## Summary

Add reverse port forwarding APIs to `AdbRunner`, enabling the MAUI DevTools CLI to manage hot-reload tunnels without going through ServiceHub.

Closes #303

## Context

Reverse port forwarding (`adb reverse`) is used across **three separate codebases** today, each with its own implementation:

1. **Visual Studio** (`ClientTools.Platform`) — raw ADB wire protocol via `TransportRunCommandWithStatus`
2. **VS Code Extension** (`vscode-maui`) — ServiceHub JSON-RPC to C# `AdbServer`
3. **Manual CLI** — no shared C# implementation

Each IDE reimplements the same ADB operations. This PR adds a **shared, CLI-based implementation** in `android-tools`.

### Types

```csharp
public enum AdbProtocol { Tcp, LocalAbstract, LocalReserved, LocalFilesystem }

public record AdbPortSpec(AdbProtocol Protocol, int Port)
{
    public string ToSocketSpec();                    // e.g. "tcp:5000"
    public static AdbPortSpec? TryParse(string spec); // parse "tcp:5000" → AdbPortSpec
}

public record AdbPortRule(AdbPortSpec Remote, AdbPortSpec Local);
```

### `AdbRunner` Methods

| Method | ADB Command | Purpose |
|--------|-------------|---------|
| `ReversePortAsync(serial, AdbPortSpec remote, AdbPortSpec local, ct)` | `adb reverse <remote> <local>` | Set up reverse forwarding |
| `RemoveReversePortAsync(serial, AdbPortSpec remote, ct)` | `adb reverse --remove <remote>` | Remove a specific rule |
| `RemoveAllReversePortsAsync(serial, ct)` | `adb reverse --remove-all` | Remove all rules |
| `ListReversePortsAsync(serial, ct)` | `adb reverse --list` | List active rules |

### Usage

```csharp
var tcp5000 = new AdbPortSpec(AdbProtocol.Tcp, 5000);

// Set up hot-reload tunnel
await adbRunner.ReversePortAsync("emulator-5554", tcp5000, tcp5000);

// List active rules
IReadOnlyList<AdbPortRule> rules = await adbRunner.ListReversePortsAsync("emulator-5554");

// Clean up
await adbRunner.RemoveReversePortAsync("emulator-5554", tcp5000);
await adbRunner.RemoveAllReversePortsAsync("emulator-5554");
```

## Key Design Decisions

[...]

PR #351, opened 2026-04-30
[AdbRunner] Add ADB forward port management (follow-up to #305)

rmarinho:
## Summary

Adds the symmetric forward-port pair to the reverse-port methods that landed in #305. Same `AdbPortSpec` / `AdbPortRule` / `AdbProtocol` types — just four new methods. Discussed in https://github.com/dotnet/android-tools/pull/305#issuecomment-4352578817.

| Method | adb command |
|--------|-------------|
| `ForwardPortAsync(serial, local, remote)` | `adb -s <serial> forward <local> <remote>` |
| `RemoveForwardPortAsync(serial, local)` | `adb -s <serial> forward --remove <local>` |
| `RemoveAllForwardPortsAsync(serial)` | `adb -s <serial> forward --remove-all` |
| `ListForwardPortsAsync(serial)` | `adb forward --list` (filtered by serial) |

## Why we need this in addition to reverse

`adb forward` and `adb reverse` are not interchangeable — they connect opposite directions:

- **`reverse <remote-on-device> <local-on-host>`** — device-side socket forwards to host-side. Used by hot reload (the device app connects to a "device" port that's actually tunnelled to the IDE host).
- **`forward <local-on-host> <remote-on-device>`** — host-side socket forwards to device-side. Used when the IDE / harness needs to connect to a service running on the device:
  - **JDWP debugger attach** — `forward tcp:N jdwp:<pid>`
  - **Performance-tracing endpoints** exposed by the runtime
  - **DevFlow agent connect** — when the in-app agent listens on a device port and the host needs a stable host-side port to reach it

## `--list` output format note

`adb forward --list` emits one line per rule across **all** devices in the form `<serial> <local> <remote>` (different from `(reverse) <remote> <local>`). `ListForwardPortsAsync` uses the unscoped `adb forward --list` and filters to the requested serial in `ParseForwardListOutput`. Serial match is case-sensitive to match adb's behaviour.

## Tests

- **12 new parser tests** (`ParseForwardListOutput_*`) mirroring the reverse parser tests:
  - single rule, multiple rules, serial filtering, empty output, no lines, empty serial, no matching device, malformed line, non-tcp specs, Windows line endings, tab separation, case-sensitivity.
- **7 new parameter-validation tests** covering empty serial / null spec for the four new public methods.

All `Forward` + `Reverse` tests pass locally (36/36). Full suite has one pre-existing unrelated `JdkDirectory_JavaHome` failure on `main` that's not touched by this PR.

## Consumers

- **VS Code MAUI extension** ServiceHub→CLI migration — `forwardPort` in `MauiAndroidPlatform.ts` (debugger configurations, perf tooling).
- **MAUI DevTools CLI** ([dotnet/maui-labs#197](https://github.com/dotnet/maui-labs/issues/197)) — `maui android port forward …` group, sibling of the existing `reverse` surface.
- **Visual Studio** `ClientTools.Platform` — same paths that drive reverse today have parallel forward call-sites.

## Public API additions

```
virtual Xamarin.Android.Tools.AdbRunner.ForwardPortAsync(string! serial, AdbPortSpec! local, AdbPortSpec! remote, CancellationToken cancellationToken = default) -> Task!
virtual Xamarin.Android.Tools.AdbRunner.ListForwardPortsAsync(string! serial, CancellationToken cancellationToken = default) -> Task<IReadOnlyList<AdbPortRule!>!>!
virtual Xamarin.Android.Tools.AdbRunner.RemoveAllForwardPortsAsync(string! serial, CancellationToken cancellationToken = default) -> Task!
virtual Xamarin.Android.Tools.AdbRunner.RemoveForwardPortAsync(string! serial, AdbPortSpec! local, CancellationToken cancellationToken = default) -> Task!
```


src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbRunner.cs
```diff
@@ -355,6 +355,126 @@ internal static IReadOnlyList<AdbPortRule> ParseReverseListOutput (IEnumerable<s
...
+	/// <param name="serial">Device serial number.</param>
+	/// <param name="local">Local (host-side) port spec.</param>
+	/// <param name="remote">Remote (device-side) port spec.</param>
+	/// <param name="cancellationToken">Cancellation token.</param>
+	public virtual async Task ForwardPortAsync (string serial, AdbPortSpec local, AdbPortSpec remote, CancellationToken cancellationToken = default)
+	{
+		if (string.IsNullOrWhiteSpace (serial))
+			throw new ArgumentException ("Serial must not be empty.", nameof (serial));
+		if (local is null)
+			throw new ArgumentNullException (nameof (local));
```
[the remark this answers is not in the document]
rmarinho: Thanks for the suggestion. Scoping forward port support to TCP-only is intentional for this PR; non-TCP socket specs (jdwp, local, etc.) are out of scope and will be addressed in a follow-up if needed.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AdbRunnerTests.cs
```diff
@@ -863,6 +863,168 @@ public void ParseReverseListOutput_TabSeparated ()
 		Assert.AreEqual (8081, rules [1].Remote.Port);
 	}
 
+	// --- ParseForwardListOutput tests ---
+	// Consumer: MAUI DevTools (via ListForwardPortsAsync — host→device tunnel for JDWP debugger
+	// attach, perf endpoints, host-side DevFlow agent connect), vscode-maui ServiceHub replacement.
```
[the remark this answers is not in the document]
rmarinho: Thanks for the suggestion. Scoping forward port support to TCP-only is intentional for this PR; non-TCP socket specs (jdwp, local, etc.) are out of scope and will be addressed in a follow-up if needed.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AdbRunnerTests.cs
```diff
@@ -1069,6 +1231,70 @@ public void ListReversePortsAsync_EmptySerial_ThrowsArgumentException ()
...
+
+	[Test]
+	public void RemoveAllForwardPortsAsync_EmptySerial_ThrowsArgumentException ()
+	{
+		var runner = new AdbRunner ("/fake/sdk/platform-tools/adb");
+		Assert.ThrowsAsync<System.ArgumentException> (
+			async () => await runner.RemoveAllForwardPortsAsync (""));
+	}
+
+	// --- ListForwardPortsAsync parameter validation tests ---
```
jonathanpeppers: Why are there region-like comments? (regions in disguise!)

The problem is when new tests are added, these cause problems when the new tests don't fit in. AIs (and humans!) will put new tests in random places.

Can we delete them all?

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbRunner.cs
```diff
@@ -355,6 +355,126 @@ internal static IReadOnlyList<AdbPortRule> ParseReverseListOutput (IEnumerable<s
...
+			throw new ArgumentNullException (nameof (remote));
+		if (local.Port <= 0 || local.Port > 65535)
+			throw new ArgumentOutOfRangeException (nameof (local), local.Port, "Port must be between 1 and 65535.");
+		if (remote.Port <= 0 || remote.Port > 65535)
+			throw new ArgumentOutOfRangeException (nameof (remote), remote.Port, "Port must be between 1 and 65535.");
+
+		var psi = ProcessUtils.CreateProcessStartInfo (adbPath, "-s", serial, "forward", local.ToSocketSpec (), remote.ToSocketSpec ());
+		using var stderr = new StringWriter ();
+		var exitCode = await ProcessUtils.StartProcess (psi, null, stderr, cancellationToken, environmentVariables).ConfigureAwait (false);
+		ProcessUtils.ThrowIfFailed (exitCode, $"adb -s {serial} forward {local} {remote}", stderr);
```
jonathanpeppers: Should we also be saving `stdout` here and pass it to `ThrowIfFailed`? Would a potential error message appear in stdout?

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbRunner.cs
```diff
@@ -355,6 +355,126 @@ internal static IReadOnlyList<AdbPortRule> ParseReverseListOutput (IEnumerable<s
...
+		if (remote is null)
+			throw new ArgumentNullException (nameof (remote));
+		if (local.Port <= 0 || local.Port > 65535)
+			throw new ArgumentOutOfRangeException (nameof (local), local.Port, "Port must be between 1 and 65535.");
+		if (remote.Port <= 0 || remote.Port > 65535)
+			throw new ArgumentOutOfRangeException (nameof (remote), remote.Port, "Port must be between 1 and 65535.");
+
+		var psi = ProcessUtils.CreateProcessStartInfo (adbPath, "-s", serial, "forward", local.ToSocketSpec (), remote.ToSocketSpec ());
+		using var stderr = new StringWriter ();
+		var exitCode = await ProcessUtils.StartProcess (psi, null, stderr, cancellationToken, environmentVariables).ConfigureAwait (false);
```
jonathanpeppers: Should we consider using the "adb tcpclient" made in another PR, or does this _need_ to shell out to a new `adb` process?

rmarinho: Good question. Here's the analysis:

**What the four forward operations need from the ADB daemon protocol:**

| `AdbRunner` method | Wire command |
| --- | --- |
| `ForwardPortAsync` | `host-serial:<serial>:forward:tcp:<local>;tcp:<remote>` → `OKAY` / `FAIL+msg` |
| `RemoveForwardPortAsync` | `host-serial:<serial>:killforward:tcp:<local>` → `OKAY` / `FAIL+msg` |
| `RemoveAllForwardPortsAsync` | `host-serial:<serial>:killforward-all` → `OKAY` / `FAIL+msg` |
| `ListForwardPortsAsync` | `host:list-forward` → `OKAY` + length-prefixed payload `<serial> <local> <remote>\n…` |

**What `AdbClient` from #327 already exposes:**
`ConnectAsync(port)`, `SendCommandAsync(string)`, `ReadStatusAsync` (`Okay`/`Fail`), `EnsureOkayAsync` (throws on `FAIL` with the daemon’s message), `ReadLengthPrefixedStringAsync`, `Reconnect`, `Dispose`. All four operations above map cleanly onto that surface with no new wire-protocol primitives needed.

**Pros of switching to `AdbClient`:**
- No process spawn per call (~50–150 ms saved on Windows; matters when build tasks call these in a loop).
- Cleaner error reporting — the daemon returns a length-prefixed FAIL message instead of us having to parse stderr/stdout text.
- Aligns with where the codebase is moving (#327, #323).

**Cons / reasons to defer:**
- **#327 is not merged.** Adopting it here couples this PR to #327’s merge order and to any shape changes still in flight on that PR.
- **Asymmetry inside `AdbRunner`.** The reverse methods merged in #305 (`ReversePortAsync`, `RemoveReversePortAsync`, `RemoveAllReversePortsAsync`, `ListReversePortsAsync`) all shell out today. Switching only the forward methods leaves the class half-and-half, which is harder to reason about than either extreme.
- **Loss of implicit `adb start-server`.** Shelling out to `adb` auto-spawns the daemon if it isn’t running. A bare socket connect to `127.0.0.1:5037` throws `SocketException`. We’d need explicit start-server logic before any `AdbClient` call.
- **Orthogonal bug worth flagging:** `host-serial:<serial>:killforward-all` is global on the daemon — it clears forwards for **all** devices, ignoring the serial. The current shell-out is the same (passes `-s <serial>` but adb does not honour it for `--remove-all`). Switching wire protocols doesn’t fix this; it’s a separate behaviour bug.

**Recommendation:** keep this PR shelling out (symmetric with the merged reverse implementation), and file a follow-up to migrate **all** of `AdbRunner` (devices, getprop, AVD name, forward, reverse) to `AdbClient` in one consistent pass once #327 has merged. That follow-up can also address the implicit-start-server question and the global `killforward-all` semantics cleanly.

Happy to take the other route if you'd rather couple this PR to #327 — let me know and I'll switch.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbRunner.cs
rmarinho: Done — ForwardPortAsync, RemoveForwardPortAsync, and RemoveAllForwardPortsAsync now capture stdout and pass it as the fourth arg to ProcessUtils.ThrowIfFailed (matching the convention used by ListForwardPortsAsync and ListReversePortsAsync). adb does sometimes write error text to stdout, so this should give us cleaner diagnostics on failure.

The non-list reverse methods (ReversePortAsync, RemoveReversePortAsync, RemoveAllReversePortsAsync) still pass only stderr today; I'd suggest the same treatment for those in a separate cleanup PR to keep this diff focused on the forward feature.

src/Xamarin.Android.Tools.AndroidSdk/Runners/AdbRunner.cs
```diff
@@ -355,6 +355,126 @@ internal static IReadOnlyList<AdbPortRule> ParseReverseListOutput (IEnumerable<s
...
+			if (parts.Length < 3)
+				continue;
+
+			if (!string.Equals (parts [0], serial, StringComparison.Ordinal))
+				continue;
+
+			var local = AdbPortSpec.TryParse (parts [1]);
+			var remote = AdbPortSpec.TryParse (parts [2]);
+			if (local is { } l && remote is { } r)
+				rules.Add (new AdbPortRule (r, l));
```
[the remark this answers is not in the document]
rmarinho: Added a <remarks> block on ParseForwardListOutput explicitly contrasting the field order with ParseReverseListOutput (forward: <serial> <local> <remote>; reverse: (reverse) <remote> <local>), plus a pointer comment at the parse site. Felt lighter than refactoring two five-line parsers into a shared helper.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AdbRunnerTests.cs
```diff
@@ -1069,6 +1231,70 @@ public void ListReversePortsAsync_EmptySerial_ThrowsArgumentException ()
...
+		Assert.ThrowsAsync<System.ArgumentException> (
+			async () => await runner.ForwardPortAsync ("", new AdbPortSpec (AdbProtocol.Tcp, 5000), new AdbPortSpec (AdbProtocol.Tcp, 5000)));
+	}
+
+	[Test]
+	public void ForwardPortAsync_NullLocal_ThrowsArgumentNull ()
+	{
+		var runner = new AdbRunner ("/fake/sdk/platform-tools/adb");
+		Assert.ThrowsAsync<System.ArgumentNullException> (
+			async () => await runner.ForwardPortAsync ("emulator-5554", (AdbPortSpec) null!, new AdbPortSpec (AdbProtocol.Tcp, 5000)));
```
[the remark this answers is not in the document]
rmarinho: Done — all three sites (lines 1249, 1257, 1275) now use (AdbPortSpec) null to match the existing reverse-port tests and the repo's no-null-forgiving-operator rule.

tests/Xamarin.Android.Tools.AndroidSdk-Tests/AdbRunnerTests.cs
rmarinho: Good call — deleted all 21 // --- ... --- divider lines in AdbRunnerTests.cs (including the pre-existing ones, since you said "delete them all"). Tests now flow as a flat list and new tests can land anywhere.

jonathanpeppers (approved):

